PR #27, opened 2023-08-09
add --update flag to bump flake inputs before rebuild

NotAShelf:
Adds `--update` so the flake inputs are bumped before the rebuild. The current implementation puts the update command *after* the --dry check, so the user will see the diff before they update the system.


viperML:
The update command should be ordered before the build command

NotAShelf:
> The update command should be ordered before the build command

That is true. I was so fixated on getting it to comply with --ask and --dry, I forgot the update needs to be done before the rebuild. Should be good now.

PR #28, opened 2023-08-11
Fix evaluation errors in nixos module

mrene:
When using the nixos module in a system that defines `system` via `nixpkgs.hostPlatform`, evaluation fails with the following error:

```
       error: Neither nixpkgs.system nor any other option in nixpkgs.* is meant
       to be read by modules and configurations.
       Use pkgs.stdenv.hostPlatform instead.
```

See a minimal reproduction flake in [this gist](https://gist.github.com/mrene/4cd68bde1af93f959f2587ae0facbe11)

Eval with: 
```
nix eval git+https://gist.github.com/4cd68bde1af93f959f2587ae0facbe11.git#nixosConfigurations.test.config.system.build.toplevel
```

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -1,6 +1,7 @@
 self: {
   config,
   lib,
+  pkgs,
   ...
 }: {
   options.nh = with lib; {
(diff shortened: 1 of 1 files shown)


viperML:
Thanks, I don't run latest nixpkgs so perhaps it was changed recently.

Anyways, why not use `pkgs.stdenv.hostPlatform` ?

mrene:
> Thanks, I don't run latest nixpkgs so perhaps it was changed recently.
> 
> Anyways, why not use `pkgs.stdenv.hostPlatform` ?

I could, I wasn't sure what the best practice was, and it seems like `pkgs.system` is always aliased.

PR #42, opened 2023-11-26
skip version diffing for fresh installs

StarrFox:
this pr fixes a crash when running `nh home switch` on a fresh nixos install (hasn't used home-manager before) because some of the home-manager paths didn't exist yet

it now just skips the version compare in these cases


viperML:
Thanks. Though I would like to implement a cleaner way to check for profiles in the classic way or xdg.

PR #47, opened 2023-12-12
add `flakeDir` option to automatically set the FLAKE variable

NotAShelf:
Adds a `flakeDir` option that sets the `FLAKE` variable for the user.


viperML:
1. Why not call it flake instead of flakeDir? If you are concerned about the user setting a flake.nix, it can be asserted.

2. Type should be types.path

3  What if the user doesn't set it? The type should be nullOr path then.

NotAShelf:
I was thinking an assertion instead of a nullOr - but that's not very backwards compatible. I'll make it `nullOr`.

NotAShelf:
Okay, it now asserts if `nh.flake` points at a file that has the `.nix` prefix.

viperML:
Checks fail

NotAShelf:
I really don't know if that mkIf is the correct way to conditionally assert.

viperML:
Why not?
```nix
assertion = (config.nh.flake != null) -> !(lib.hasSuffix ".nix" config.nh.flake)
```

NotAShelf:
Mainly because I didn't know how the `->` operator worked. Will change.

viperML:
Is it ready?

NotAShelf:
Will commit tonight.

NotAShelf:
Should be good now.

Issue #46, opened 2023-12-08
Fix `nh os <something> --update`


viperML:
https://github.com/viperML/nh/blob/375c6cf57de3a839b7937358659bea526da27eae/src/nixos.rs#L54

Nix 2.19 changes the syntax

eclairevoyant:
context: https://github.com/NixOS/nix/blob/master/src/nix/flake-update.md

`nix flake update (arg)` now updates the flake input named (arg); updating a flakeref requires `--flake (flakeref)`

eclairevoyant:
#75 fixes this?

NotAShelf:
'puter... my worst enemy...

PR #65, opened 2024-01-28
fix --update on newer nix versions

NotAShelf:
The current iteration of the `--update` flag is missing `--flake` as described by `nix flake update --help`. This MR adds it (big surprise.)

Currently a draft because I'd like to implement a version gate for those below v2.19 instead of failing spectacularly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -768,7 +768,7 @@ dependencies = [
 
 [[package]]
 name = "nh"
-version = "3.5.5"
+version = "3.5.6"
 dependencies = [
  "ambassador",
  "anstyle",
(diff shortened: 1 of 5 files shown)


viperML:
related: #46

viperML:
As I said on the issue, I'd think it would be fine for nixos-unstable's nix to reach 2.19 and then just don't check the version.

NotAShelf:
Does it make sense to implement a version gate, or should we actually wait for 2.19 to drop?

NotAShelf:
Come to think of it, what about people who use lower Nix versions on purpose? I think we should be handling it *somehow* instead of the command failing silently in the background.

NotAShelf:
I forgot about this

NotAShelf:
I was adding a check :(

PR #74, opened 2024-03-08
add: build-vm command to os subcommand

xiro-codes:
Implements build-vm command into nh

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -85,6 +85,8 @@ pub enum OsRebuildType {
     Boot(OsRebuildArgs),
     /// Build and activate the new configuration
     Test(OsRebuildArgs),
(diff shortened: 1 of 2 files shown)


viperML:
Couldn't this be retrofitted into OsRebuildArgs::rebuild instead?

xiro-codes:
Ill look into it, also love you blog it really helped me get into nix early on.

viperML:
So what do you want to do with this command? It is named `build-vm`, but it actually runs the vm. Also it shouldn't install it into the system profile, and probably neither ask for confirmation or diff.

xiro-codes:
I needed to test my config that was broken. so i added it didn't think much farther than that.
ps. it was still broken but its fixed now

PR #82, opened 2024-04-08
fix: "nh os test" do not set nix-profile

### LINKED ISSUE #81: `nh os test` builds are being added to the bootloader menu afterwards
`nh os test` sets nix-profile by running `nix-env --profile --set`. 
This does not affect the current build, but the following `nh os switch` adds this test build to the bootloader menu.

psfloyd:
The new behaviour is consistent with "nixos-rebuild test". The nix-profile is only set on "boot" and "switch".
In the old implementation, builds with "nh os test" were later added to the boot menu on "nh os switch".

Fixes #81.

### DIFF
--- a/src/nixos.rs
+++ b/src/nixos.rs
@@ -126,18 +126,6 @@ impl OsRebuildArgs {
             }
         }
 
-        commands::CommandBuilder::default()
-            .args([
-                "sudo",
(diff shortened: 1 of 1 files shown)


viperML:
Thanks

PR #84, opened 2024-04-17
fix: don't clean gcroots if --nogcroots is passed

thrombe:
`nh clean <command> --nogcroots` does not ignore any of the gcroots.

I noticed that the [nogcroots](https://github.com/viperML/nh/blob/87ede1ab13beb752b94c23a7b3d03c6b533fa5cb/src/interface.rs#L200) flag is never actually used.

I implemented it such that the gcroots detected by nh are still shown.
image

or would it be better to not detect those if `--nogcroots` is passed?

### DIFF
--- a/src/clean.rs
+++ b/src/clean.rs
@@ -148,7 +148,11 @@ impl NHRunnable for interface::CleanMode {
                             gcroots_tagged.insert(dst, false);
                         }
                         Ok(_) => {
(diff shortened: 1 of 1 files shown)


viperML:
At some point `--nogcroots` did things but got lost in a refactor. Thanks for the heads-up

Issue #59, opened 2024-01-04
Use pkexec or doas


viperML:
Optionally elevate to root with pkexec or doas.

Ideally clean-up how to self-elevate, instead of calling `sudo argv[..]`

PR #92, opened 2024-04-20
fix: support to other privilege elevation programs.

henriquekirchheck:
This pr adds support for both `doas` and `pkexec`, while keeping support for `sudo`.

For this to work, I decided that the cleanest way to implement this was by adding a extra parameter to the Command struct that specifies if the command should be ran as root or not. When the root parameter is set, it checks for the existence of `doas`, `sudo`, and `pkexec`, in that order, and uses it to run the command by prefixing the command invocation arguments with the respective program when it is constructed.

The specified order was chosen because I believe that if someone has `doas` installed in their system, it's more likely that they use it as their main privilege escalation program. ´pkexec´ was added more as a fallback in case someone doesn't have `doas` or `sudo`.

This was tested in my NixOS Unstable system with both `sudo` and `doas`, but I haven't tested `pkexec` personally.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -9,6 +9,8 @@ use thiserror::Error;
 use subprocess::{Exec, ExitStatus, Redirection};
 use tracing::{debug, info};
 
+use crate::util::get_elevation_program;
+
 #[derive(Debug, derive_builder::Builder)]
 #[builder(derive(Debug), setter(into))]
 pub struct Command {
@@ -21,6 +23,9 @@ pub struct Command {
     /// Arguments 0..N
     #[builder(setter(custom))]
     args: Vec<OsString>,
+    /// Whether to run the command as root or not
+    #[builder(default = "false")]
+    root: bool,
 }
 
 impl CommandBuilder {
@@ -38,9 +43,7 @@ impl CommandBuilder {
 
 impl Command {
     pub fn exec(&self) -> Result<()> {
-        let [head, tail @ ..] = &*self.args else {
-            bail!("Args was length 0");
-        };
+        let (head, tail) = self.get_cmd_head_args()?;
 
         let cmd = Exec::cmd(head)
             .args(tail)
@@ -64,9 +67,7 @@ impl Command {
     }
 
     pub fn exec_capture(&self) -> Result<Option<String>> {
-        let [head, tail @ ..] = &*self.args else {
-            bail!("Args was length 0");
-        };
+        let (head, tail) = self.get_cmd_head_args()?;
 
         let cmd = Exec::cmd(head)
             .args(tail)
@@ -82,6 +83,17 @@ impl Command {
             Ok(Some(cmd.capture()?.stdout_str()))
         } else {
             Ok(None)
+        }
+    }
+
+    fn get_cmd_head_args<'a>(&'a self) -> Result<(OsString, &'a [OsString])> {
+        if self.root {
+            Ok((get_elevation_program()?, self.args.as_ref()))
+        } else {
+            let [head, tail @ ..] = &*self.args else {
+                bail!("Args was length 0");
+            };
+            Ok((head.clone(), tail))
         }
     }
 }
--- a/src/util.rs
+++ b/src/util.rs
@@ -2,7 +2,10 @@ extern crate semver;
 
 use color_eyre::{eyre, Result};
 use semver::Version;
+use tracing::debug;
+use which::which;
 
+use std::ffi::OsString;
 use std::process::Command;
 use std::str;
 
@@ -55,4 +58,43 @@ pub fn get_nix_version() -> Result<String> {
     }
 
     Err(eyre::eyre!("Failed to extract version"))
+}
+
+/// Gets a path to a previlege elevation program based on what is available in the system.
+///
+/// This funtion checks for the existence of common privilege elevation program names in
+/// the `PATH` using the `which` crate and returns a Ok result with the `OsString` of the
+/// path to the binary. In the case none of the checked programs are found a Err result is returned.
+///
+/// The search is done in this order:
+///
+/// 1. `doas`
+/// 1. `sudo`
+/// 1. `pkexec`
+///
+/// The logic for choosing this order is that a person with doas installed is more likely to be
+/// using it as their main privilege elevation program.
+///
+/// # Returns
+///
+/// * `Result<OsString>` - The absolute path to the privilege elevation program binary or an error if a
+/// program can't be found.
+pub fn get_elevation_program() -> Result<OsString> {
+    let has_doas = which("doas");
+    if let Ok(path) = has_doas {
+        debug!(?path, "doas path found");
+        return Ok(path.into_os_string());
+    }
+    let has_sudo = which("sudo");
+    if let Ok(path) = has_sudo {
+        debug!(?path, "sudo path found");
+        return Ok(path.into_os_string());
+    }
+    let has_pkexec = which("pkexec");
+    if let Ok(path) = has_pkexec {
+        debug!(?path, "pkexec path found");
+        return Ok(path.into_os_string());
+    }
+
+    Err(eyre::eyre!("No elevation strategy found"))
 }
--- a/src/util.rs
+++ b/src/util.rs
@@ -6,7 +6,6 @@ use tracing::debug;
 use which::which;
 
 use std::ffi::OsString;
-use std::path::PathBuf;
 use std::process::Command;
 use std::str;
 
@@ -79,6 +78,7 @@ pub fn get_nix_version() -> Result<String> {
 ///
 /// 1. `doas`
 /// 1. `sudo`
+/// 1. `run0`
 /// 1. `pkexec`
 ///
 /// The logic for choosing this order is that a person with doas installed is more likely
@@ -91,11 +91,14 @@ pub fn get_nix_version() -> Result<String> {
 pub fn get_elevation_program() -> Result<OsString> {
     let args = <NHParser as clap::Parser>::parse();
     if let Some(path) = args.elevation_program.map(|path| which(path)) {
-        debug!(?path, "privilege elevation path specified via command line argument");
+        debug!(
+            ?path,
+            "privilege elevation path specified via command line argument"
+        );
         return Ok(path?.into_os_string());
     }
 
-    const STRATEGIES: [&str; 3] = ["doas", "sudo", "pkexec"];
+    const STRATEGIES: [&str; 4] = ["doas", "sudo", "run0", "pkexec"];
 
     for strategy in STRATEGIES {
         if let Ok(path) = which(strategy) {
--- a/src/util.rs
+++ b/src/util.rs
@@ -6,8 +6,11 @@ use tracing::debug;
 use which::which;
 
 use std::ffi::OsString;
+use std::path::PathBuf;
 use std::process::Command;
 use std::str;
+
+use crate::interface::NHParser;
 
 /// Compares two semantic versions and returns their order.
 ///
@@ -60,11 +63,17 @@ pub fn get_nix_version() -> Result<String> {
     Err(eyre::eyre!("Failed to extract version"))
 }
 
-/// Gets a path to a previlege elevation program based on what is available in the system.
+/// Gets a path to a previlege elevation program based on what is available in the system
+/// or from the command line arguments.
 ///
-/// This funtion checks for the existence of common privilege elevation program names in
-/// the `PATH` using the `which` crate and returns a Ok result with the `OsString` of the
-/// path to the binary. In the case none of the checked programs are found a Err result is returned.
+/// First this function checks for a program specified by the command line arguments,
+/// checks if it exists in the `PATH` with the `which` crate and returns a Ok result
+/// with the OsString of the path to the binary or a Err result if not.
+///
+/// If no command line arguments were found, this funtion checks for the existence of
+/// common privilege elevation program names in the `PATH` using the `which` crate and
+/// returns a Ok result with the `OsString` of the path to the binary. In the case none
+/// of the checked programs are found a Err result is returned.
 ///
 /// The search is done in this order:
 ///
@@ -72,17 +81,23 @@ pub fn get_nix_version() -> Result<String> {
 /// 1. `sudo`
 /// 1. `pkexec`
 ///
-/// The logic for choosing this order is that a person with doas installed is more likely to be
-/// using it as their main privilege elevation program.
+/// The logic for choosing this order is that a person with doas installed is more likely
+/// to be using it as their main privilege elevation program.
 ///
 /// # Returns
 ///
 /// * `Result<OsString>` - The absolute path to the privilege elevation program binary or an error if a
 /// program can't be found.
 pub fn get_elevation_program() -> Result<OsString> {
-    let strategies = ["doas", "sudo", "pkexec"];
+    let args = <NHParser as clap::Parser>::parse();
+    if let Some(path) = args.elevation_program.map(|path| which(path)) {
+        debug!(?path, "privilege elevation path specified via command line argument");
+        return Ok(path?.into_os_string());
+    }
 
-    for strategy in strategies {
+    const STRATEGIES: [&str; 3] = ["doas", "sudo", "pkexec"];
+
+    for strategy in STRATEGIES {
         if let Ok(path) = which(strategy) {
             debug!(?path, "{strategy} path found");
             return Ok(path.into_os_string());
--- a/src/util.rs
+++ b/src/util.rs
@@ -80,20 +80,13 @@ pub fn get_nix_version() -> Result<String> {
 /// * `Result<OsString>` - The absolute path to the privilege elevation program binary or an error if a
 /// program can't be found.
 pub fn get_elevation_program() -> Result<OsString> {
-    let has_doas = which("doas");
-    if let Ok(path) = has_doas {
-        debug!(?path, "doas path found");
-        return Ok(path.into_os_string());
-    }
-    let has_sudo = which("sudo");
-    if let Ok(path) = has_sudo {
-        debug!(?path, "sudo path found");
-        return Ok(path.into_os_string());
-    }
-    let has_pkexec = which("pkexec");
-    if let Ok(path) = has_pkexec {
-        debug!(?path, "pkexec path found");
-        return Ok(path.into_os_string());
+    let strategies = ["doas", "sudo", "pkexec"];
+
+    for strategy in strategies {
+        if let Ok(path) = which(strategy) {
+            debug!(?path, "{strategy} path found");
+            return Ok(path.into_os_string());
+        }
     }
 
     Err(eyre::eyre!("No elevation strategy found"))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -474,6 +474,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d231dfb89cfffdbc30e7fc41579ed6066ad03abda9e567ccafae602b97ec5024"
 
 [[package]]
+name = "home"
+version = "0.5.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e3d1354bf6b7235cb4a0576c2619fd4ed18183f689b12b006a0ee7329eeff9a5"
+dependencies = [
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "hostname"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -753,6 +762,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "uzers",
+ "which",
 ]
 
 [[package]]
@@ -1603,6 +1613,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "which"
+version = "6.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8211e4f58a2b2805adfbefbc07bab82958fc91e3836339b1ab7ae32465dce0d7"
+dependencies = [
+ "either",
+ "home",
+ "rustix 0.38.32",
+ "winsafe",
+]
+
+[[package]]
 name = "winapi"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1791,6 +1813,12 @@ dependencies = [
  "cfg-if",
  "windows-sys 0.48.0",
 ]
+
+[[package]]
+name = "winsafe"
+version = "0.0.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d135d17ab770252ad95e9a872d365cf3090e3be864a34ab46f48555993efc904"
 
 [[package]]
 name = "zeroize"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,3 +53,4 @@ tracing-subscriber = { version = "0.3.18", features = [
     "std"
 ] }
 uzers = { version = "0.11.3", default-features = false }
+which = "6.0.1"
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -47,6 +47,10 @@ pub struct NHParser {
     /// Show debug logs
     pub verbose: bool,
 
+    #[arg(short, long, global = true, value_hint = clap::ValueHint::ExecutablePath)]
+    /// Choose what privilege elevation program should be used
+    pub elevation_program: Option<OsString>,
+
     #[command(subcommand)]
     pub command: NHCommand,
 }
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -47,7 +47,7 @@ pub struct NHParser {
     /// Show debug logs
     pub verbose: bool,
 
-    #[arg(short, long, global = true, value_hint = clap::ValueHint::ExecutablePath)]
+    #[arg(short, long, global = true, env, value_hint = clap::ValueHint::ExecutablePath)]
     /// Choose what privilege elevation program should be used
     pub elevation_program: Option<OsString>,
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,6 +10,7 @@ mod util;
 
 use crate::interface::NHParser;
 use crate::interface::NHRunnable;
+use crate::util::get_elevation_program;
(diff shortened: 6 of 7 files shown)


henriquekirchheck:
Forgot to mention that it will Fix #59, but it still uses `sudo argv[..]`

Makesesama:
I also tested on NixOs Unstable with doas installed. Works fine for me.

henriquekirchheck:
I'll look into the clean command and fix it

henriquekirchheck:
Seems like the self elevate function is only using `sudo` but I should be able to just replace it with my `get_elevation_program` function. 

@Makesesama If you could test it I would be very grateful.

Makesesama:
Seems to work now

henriquekirchheck:
Thank you!

arthsmn:
You might also be interested in adding run0 support.

Related post: https://mastodon.social/@pid_eins/112353324518585654

henriquekirchheck:
systemd is currently at version 255.4 in NixOS unstable so I'll refrain to add it until it is actually released, but I'll look into it when it is available

src/util.rs
```diff
@@ -56,3 +59,42 @@ pub fn get_nix_version() -> Result<String> {
...
+/// 1. `pkexec`
+///
+/// The logic for choosing this order is that a person with doas installed is more likely to be
+/// using it as their main privilege elevation program.
+///
+/// # Returns
+///
+/// * `Result<OsString>` - The absolute path to the privilege elevation program binary or an error if a
+/// program can't be found.
+pub fn get_elevation_program() -> Result<OsString> {
```
NotAShelf: Is this large if block necessary? I would probably do something like 

```rust
pub fn get_elevation_program() -> Result<OsString> {
    let strategies = vec!["doas", "sudo", "pkexec"];

    for strategy in strategies {
        if let Ok(path) = which(strategy) {
            debug!(?path, "{} path found", strategy);
            return Ok(path.into_os_string());
        }
    }

    Err(eyre::eyre!("No elevation strategy found"))
}
```

henriquekirchheck: That makes sense, I will update the code shortly. Only change I would make is to remove the Vec allocation and use a simple array.

```rust
pub fn get_elevation_program() -> Result<OsString> {
    let strategies = ["doas", "sudo", "pkexec"];

    for strategy in strategies {
        if let Ok(path) = which(strategy) {
            debug!(?path, "{strategy} path found");
            return Ok(path.into_os_string());
        }
    }

    Err(eyre::eyre!("No elevation strategy found"))
}
```

boomshroom:
An alternative is to allow for some way for the user to specify what privilege escalator to use. This would be easier if `nh`  had a config file though.

This would allow for forcing `pkexec` even if `sudo` is present, or using `run0` (or its predecessor `systemd-run`) without needing to update `nh` to support it.

deviantsemicolon:
Now that systemd v256 is out and currently being reviewed for eventual merging into nixpkgs, will this be updated to add support for run0

henriquekirchheck:
I'll update everything with the suggestions, thank you!

henriquekirchheck:
I guess I could add `run0` to the list of automatically checked programs, because even if it isn't released yet in nixpkgs unstable it should be fine for people that don't have it because it will just be ignored by the checks.

Now I'm not sure where in that list it should be included, considering that it will be installed in every systemd based system, I think it should be low in the list, maybe above `pkexec`. I'm open for second opinions.

Also, it should probably check for a environment variable to determine what to do for convenience, just like `nh` already does with the `FLAKE` environment variable, but I don't know of any standard environment variable names that specify that.

henriquekirchheck:
> And of course it would be ideal in such a scenario to be able to include extra arguments in case the user had particular needs (would be necessary for `systemd-run` since it doesn't pass through stdio without either `-t` or `-P`).

Any ideas on how that would look like in the command line? I added a way to specify a custom privilege elevation binary to use, but I cannot think of a way to add the extra command line arguments. Maybe another flag would work, but that seems messy to me.

A config file would be optimal, both for the `FLAKE` environment variable and the custom privilege elevation program, but I'd probably wait for @viperML to comment on what he wants to do.

deviantsemicolon:
Systemd 256 was merged into staging-next and should be in master and unstable very soon

henriquekirchheck:
@deviantsemicolon run0 support was already added! Testing would be appreciated.

deviantsemicolon:
> @deviantsemicolon run0 support was already added! Testing would be appreciated.

If they ever get around to merging the next staging-next thing, where the systemd 256 PR is, into master, then I might be able to see what I can do

deviantsemicolon:
@henriquekirchheck i disabled sudo and ran nix run to test your code. I got an error regarding setting a locale and it asked me to reauthenticate a few times, but that could be a nix run issue. not sure though. Other than that, it did use run0, so LGTM

deviantsemicolon (approved):
@henriquekirchheck i disabled sudo and ran nix run to test your code. I got an error regarding setting a locale and it asked me to reauthenticate a few times, but that could be a nix run issue. not sure though. Other than that, it did use run0 as expected, so LGTM

the error basically said it couldn't set a locale and fell back to a standard one called "C"

henriquekirchheck:
@deviantsemicolon this issue is normally related to localization, and considering my code is basically just calling a different program it doesn't feel related. I'm not currently able to test it myself, but I'll try to do it asap.

calebstewart:
I'm also getting the localization warnings mentioned above. My best guess is it's related to the environment differences w/ `run0` which exposes little to nothing of the calling environment to the subprocess. To be a bit more clear, I'll add the exact warnings I'm getting below. They are warnings from Perl, but they don't appear to be causing any actual problems. The build and install appears to have worked properly.

The only real issue I see is that w/ `run0`, there is no cached authentication, so you are prompted 3 times during a run of `nh os switch` with `run0`. :/

```
❯ nh os switch -e run0
> Building NixOS configuration
warning: Git tree '/home/caleb/git/nix' is dirty
warning: input 'nixos-hardware' has an override for a non-existent input 'nixpkgs'
Finished at 12:37:43 after 0s
> Comparing changes
<<< /run/current-system
>>> /tmp/nh-os-nuyqlh/result
No version or selection state changes.
Closure size: 2610 -> 2610 (0 paths added, 0 paths removed, delta +0, disk usage +0B).
> Activating configuration
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
activating the configuration...
setting up /etc...
reloading user units for caleb...
restarting sysinit-reactivation.target
> Adding configuration to bootloader
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```

On the positive side, the automatic coloring of output from the elevated subprocesses from `run0` is pretty nifty. It's really easy to see what logs are from elevated sections.

image

viperML:
You can use any elevation program with `run0 nh os switch -R`, and a cleaner abstraction of `--use-remote-sudo` will be implemented in #107

deviantsemicolon:
> You can use any elevation program with `run0 nh os switch -R`, and a cleaner abstraction of `--use-remote-sudo` will be implemented in #107

I would like the ability to set a default in settings so I don't need to type run0 and -R everytime. 

It also does not work on my system, and outputs an error saying it can't find the NixOS configuration, even when specifying the directory my flake is in.

viperML:
Please send the actual log with `--verbose`

deviantsemicolon:
> Please send the actual log with `--verbose`

```
TRACE nh::logging:86: Logging OK
DEBUG nh:22: args=NHParser { verbose: true, command: Os(OsArgs { action: Switch(OsRebuildArgs { common: CommonRebuildArgs { dry: false, ask: false, flakeref: FlakeRef("./flake.nix"), update: false, no_nom: false, diff_provider: "nvd diff", out_link: None }, hostname: None, specialisation: None, no_specialisation: false, extra_args: [], bypass_root_check: true }) }) }
DEBUG nh:23: NH_VERSION=3.5.26
! Bypassing root check, now running nix as root @ src/nixos.rs:32
DEBUG nh::nixos:58: out_path=("/tmp/nh-osBMrumc/result", TempDir { path: "/tmp/nh-osBMrumc" })
> Building NixOS configuration
DEBUG nh::commands:136: cmd=Pipeline { nix build './flake.nix#nixosConfigurations.""nsa-laptop"".config.system.build.toplevel' --log-format internal-json --verbose --out-link /tmp/nh-osBMrumc/result | nom --json }
Error:
   0: Building NixOS configuration
   1: No such file or directory (os error 2)
   2: No such file or directory (os error 2)

Location:
   src/commands.rs:149
```

deviantsemicolon:
this worked with the PR perfectly fine.

henriquekirchheck:
> You can use any elevation program with `run0 nh os switch -R`, and a cleaner abstraction of `--use-remote-sudo` will be implemented in #107

I think that `--use-remote-sudo` is a better way to do this, so I don`t have any problems

henriquekirchheck:
So, I've recently been contemplating the idea of re-opening this PR.  
I didn't care much for the closing of the PR at first due to the `--use-remote-sudo` flag, which fixed the problem I was personally having, however it doesn't fix the underlying problem of it still only using `sudo`.  

There has been a few issues and PRs opened with the goal of supporting other programs other than `sudo`, notably:

- #236 
- #339 
- #240

And looking at proposed solutions I still think that the one I proposed with this PR is the most elegant (I'm however open to implement new ideas or respond to feedback), and therefore would like to have this reopened.

I'll get back and rebase this my branch into master if the PR is reopened to make it suitable for a merge.

Pinging @NotAShelf because he's been active in the issues mentioned.

NotAShelf:
I'm having a difficult time putting my thoughts into words so I'll give you the short answer: **feel free to re-open this PR**. I'm going to try and review it when I get the chance. My time is very limited at the moment, and you might've noticed that the codebase is due a good cleanup. Naturally I'd prefer to allocate however limited time I have to the cleanup, but I promise to make some time for your PR if you decide to re-open it.

Do note, I'd *really* like to consolidate some of the repeated code across platform-specific modules. Such a change might create difficult merge conflicts to resolve but I can hold it off for a while if you are interested in getting this PR back to speed. *If* we can get a good, generic-enough implementation that will not be too difficult for me to manage, then I'm happy to review and possibly merge. It appears to me that viper is currently taking a break (and I have no way of communicating him), which leaves me as the one to make the executive decision. In which case I say go for it.

NotAShelf:
Me: "short answer"
Also me: an entire paragraph

Okay well you get the point. I'm happy to try getting more escalation utilities into nh, since that has been a long-time request.

henriquekirchheck:
Alright, I'll get into updating this, but I understand the concern with making it more difficult to refactor down the road. I'll try my best to make this as generic and self-contained as possible to make changing it in the future as simple as possible.

Also, I don't think I have the permission to reopen this PR by myself as I wasn't the one that closed it, but I'll start work on it ASAP.

henriquekirchheck:
The main difficulty I see with supporting other programs is the lack of support for --preserve-env, which of the four I pretend to support, only `run0` and `sudo` supports it.

The way that nixos-rebuild-ng seems to be doing it is just passing all args via `sudo env {ARG=VALUE...} [COMMAND] {ARGS...}` when running via ssh, but it also doesn't inherit from the current env so it's functionally equivalent to the current implementation.

But in general, it seems like they don't need any extra env that `sudo` doesn't automatically preserve, but I haven't studied this code base enough to know why these extra envs are needed.

Now, I think the way to progress in a compatible way is to just read the current value of the variable that is meant to be preserved and just add it to the `env` args, which would be pretty straightforward and reduce the logic needed to support extra programs in the future if they come to exist or even to make sure that a custom program that is passed in can be used instead of the officially supported options.

I will be progressing with this idea, however I don't know if it could cause any problems because of the use of `env::set_var` throughout the code base. But I assume it wont considering I couldn't find any use of threads in the affected paths.

hqnna:
> The main difficulty I see with supporting other programs is the lack of support for --preserve-env, which of the four I pretend to support, only `run0` and `sudo` supports it.

This is not correct. You can configure doas to support keeping env similarly via the config:
https://man.openbsd.org/doas.conf#keepenv

To do it you could just assign a custom rule to the group you want:
```nix
security.doas.extraRules = [{
  groups = [ "wheel" ];
  keepenv = true;
}];
```

It is via a config option rather than a flag, but it does support preserving env.

henriquekirchheck:
> This is not correct. You can configure doas to support keeping env similarly via the config: https://man.openbsd.org/doas.conf#keepenv

I'm aware of this option, however it is not dynamic and doesn't seem to have a option to filter some specific variables. I still think the `{elevation} env VAR=VAL {command}` is better as it should be generic for any privilege elevation programs, however I'll look into it in the case I can't make it work.

NotAShelf:
Sorry for the constant changes to relevant modules. There were some things I wanted to get out of my way for the 4.2.0 release. Should be around the last time I've created merge conflicts, maybe one or two more times with the SSH changes I'm planning.

Once the master changes are done, I'd like to review and potentially merge this PR. Perhaps even in time for 4.2.0.

src/util.rs
```diff
@@ -275,3 +271,51 @@ pub fn print_dix_diff(old_generation: &Path, new_generation: &Path) -> Result<()
...
+    let args = <Main as clap::Parser>::parse();
+    if let Some(path) = args.elevation_program.map(|path| which(path)) {
+        debug!(
+            ?path,
+            "privilege elevation path specified via command line argument"
+        );
+        return Ok(path?.into_os_string());
+    }
+
+    const STRATEGIES: [&str; 4] = ["doas", "sudo", "run0", "pkexec"];
```
midischwarz12: Could also add `su -c` or whatever it is as a last resort strategy. I'd add this to the back of the list. You'd have to modify your logic to add the command line flag if `su` though. 

I'd also put `run0` and `pkexec` ahead of `sudo` considering many that use `run0` or `pkexec` may also have `sudo` but want to refrain from using it.

boomshroom: `run0` is packaged with systemd, so it will pretty much always be present on NixOS systems. `sudo` meanwhile is its own separate option that can easily be disabled, or uninstalled on other systems. `pkexec` is enabled through polkit, which is also required for `run0`. Ordering `pkexec` vs `run0` I don't know which would be best, but setting `security.sudo.enable = false` is a lot easier than trying to disable `run0` or trying to disable the wrapper for `pkexec`, removing the unwrapped version from the `PATH`, and preserving the wrapper for `polkit-agent-helper-1`.

midischwarz12: Alright, fair enough.

henriquekirchheck:
This should be working as expected, it creates a `nh::commands::ElevationStrategy` based on the command line arguments and propagates that to the commands that need it.

This was tested on `nixos-unstable` with `doas`, "`sudo`" (an alias to doas), `run0`, and `pkexec`. I don't have a Darwin system, so I can't test for problems there, but it should work.

Any feedback appreciated.

henriquekirchheck:
Re-implemented `NH_SUDO_PRESERVE_ENV` as `NH_PRESERVE_ENV`. It should work the same, but now it just omits the environment variables from the `env` arguments instead of not passing it to `sudo`

henriquekirchheck:
Just as a heads up, unless there is any change needed in the code or some bug I couldn't catch, I think this should be complete and ready for merging. Please let me know if any changes are desired, be it something in the choice/environment logic or tests or whatever else.

NotAShelf:
Sorry for the radio silence, I've been a bit too swamped at work and what little time I've got has been going into personal projects & additions that I will forget otherwise. I'll provide you a review later today, if you get a chance to deal with the merge conflicts by then but I'll try to fix those myself if I end up merging this today. The implementation looks good from a first glance anyway, so I wouldn't expect any substantial changes.

midischwarz12:
I'll have quite a bit of time to test this tomorrow. Feel free to ping me if I forget.

src/commands.rs
```diff
@@ -221,63 +288,44 @@ impl Command {
         cmd
     }
 
+    /// Creates a Exec that contains elevates the program with proper environment handling.
+    ///
+    /// Panics: If called when `self.elevate` is `None`
     fn build_sudo_cmd(&self) -> Exec {
-        let mut cmd = Exec::cmd("sudo");
+        let elevation_program = self.elevate.as_ref().unwrap().resolve().unwrap();
```
NotAShelf: I don't think I like this. Very unsafe and in my opinion not very suitable for production. Could you perhaps return `Result<Exec>` and handle errors properly?

E.g.,

```rust
let elevation_program = self.elevate
    .as_ref()
    .ok_or_else(|| eyre::eyre!("Command not found for elevation"))?
    .resolve()
    .context("Failed to resolve elevation program")?;
```

src/commands.rs
```diff
@@ -221,63 +288,44 @@ impl Command {
...
-                    ]);
-                }
-                _ => {
-                    cmd = cmd.args(&[
-                        "--set-home",
-                        &format!("--preserve-env={}", preserve_vars.join(",")),
-                    ]);
-                }
+        // Use NH_SUDO_ASKPASS program for sudo if present
+        if elevation_program.to_string_lossy().contains("sudo") {
```
NotAShelf: This is brittle and will break with programs like "pseudo" or paths containing "sudo" in directory names. You should be using proper program name extraction. Maybe something along the lines of:

```rust
let program_name = elevation_program
    .as_ref()
    .file_name()
    .and_then(|name| name.to_str())
    .unwrap_or("");
if program_name == "sudo" { ... }
```

src/commands.rs
```diff
@@ -36,13 +37,79 @@ pub enum EnvAction {
...
+    /// `run0` and `pkexec` are preinstalled in any `NixOS` system with polkit support installed,
+    /// so they have been placed lower as it's easier to deactivate sudo than it is to remove
+    /// `run0`/`pkexec`
+    ///
+    /// # Returns
+    ///
+    /// * `Result<OsString>` - The absolute path to the privilege elevation program binary or an error if a
+    /// program can't be found.
+    fn choice() -> Result<OsString> {
+        const STRATEGIES: [&str; 4] = ["doas", "sudo", "run0", "pkexec"];
```
NotAShelf, line 95: run0 requires systemd 256+ (?) and shouldn't precede pkexec, which is more universally available.

src/commands.rs
```diff
@@ -36,13 +37,79 @@ pub enum EnvAction {
...
+
+        for strategy in STRATEGIES {
+            if let Ok(path) = which(strategy) {
+                debug!(?path, "{strategy} path found");
+                return Ok(path.into_os_string());
+            }
+        }
+
+        Err(eyre::eyre!("No elevation strategy found"))
```
NotAShelf:
```suggestion
        Err(eyre::eyre!("No elevation strategy found. Checked: {}", STRATEGIES.join(", ")))
```

Or similar, for the sake of clarity.

src/commands.rs
```diff
@@ -36,13 +37,79 @@ pub enum EnvAction {
...
+    /// the `PATH` using the `which` crate and returns a Ok result with the `OsString` of the
+    /// path to the binary. In the case none of the checked programs are found a Err result is
+    /// returned.
+    ///
+    /// The search is done in this order:
+    ///
+    /// 1. `doas`
+    /// 1. `sudo`
+    /// 1. `run0`
+    /// 1. `pkexec`
```
NotAShelf:
```suggestion
    /// 1. `doas`
    /// 2. `sudo`
    /// 3. `run0`
    /// 4. `pkexec`
```

Formatter issue?

src/commands.rs
```diff
@@ -221,63 +288,44 @@ impl Command {
...
-        } else if cfg!(target_os = "macos") {
-            cmd = cmd.arg("--set-home");
         }
 
-        // Use NH_SUDO_ASKPASS program for sudo if present
-        if let Ok(askpass) = std::env::var("NH_SUDO_ASKPASS") {
-            cmd = cmd.env("SUDO_ASKPASS", askpass).arg("-A");
-        }
+        // NH_PRESERVE_ENV: set to "0" to disable preserving environment variables, "1" to force, unset defaults to force
+        let preserve_env = std::env::var("NH_PRESERVE_ENV")
```
NotAShelf, line 321: Clap should be able to handle this with `value_parser = clap::builder::BoolishValueParser::new())` if I'm not mistaken. Should help you avoid manual parsing.

NotAShelf (changes requested):
Few changes for the sake of clean code and cleaner error handling. I've been working on getting rid of all unwraps and panic-based error handling for the past few weeks.

src/commands.rs
```diff
@@ -221,63 +288,44 @@ impl Command {
...
+            .as_deref()
+            .map(|x| match x {
+                "0" => false,
+                "1" | _ => true,
```
faukah:
```suggestion
                 _ => true,
```

src/commands.rs
henriquekirchheck: Yeah, that makes sense. My logic was that if a operation needed elevated privileges and the system didn't have one it would make it a unrecoverable error and therefore a panic would be "acceptable". I'll create a Result and propagate it up.

henriquekirchheck: I'll still be unwrapping in the tests, as it should always be valid due to the `ElevationStrategy::Force`

src/commands.rs
midischwarz12: @NotAShelf If <systemd 256, it would fall back to pkexec. If pkexec precedes run0, it'll never get used despite existing.

src/commands.rs
midischwarz12: @NotAShelf It is part of the original markdown standard and all other flavors that you can just use `1.` for ordered list. When rendered, it'll automatically increment.

NotAShelf: Oh that's neat, didn't know Rustdoc format supported it.

midischwarz12: It should since it used the CommonMark specification.

src/commands.rs
henriquekirchheck: Seems to work!

src/commands.rs
henriquekirchheck: I choose run0 to be first because it's less universally available. I think of it like this:

- If someone has doas, they probably want to use doas
- If someone has sudo, they probably want to use sudo
- If someone has neither, they probably want to use run0, unless they are using systemd <=255, where in that case we use pkexec as a fallback because of how universally available it is.

src/commands.rs
henriquekirchheck: `clap::builder::BoolishValueParser` doesn't seem to have a way to parse a standalone string outside of the larger command line parser. 
I'd have to add it to the `interface::Main` parser and propagate it down because it would be a global argument, which seems overkill for this single use of the environment variable.

src/commands.rs
midischwarz12: Right, you start with the more specific cases and then work your way down to more universal or general cases as fall backs. If you start with a universal case, none of the specific cases will ever match.

src/commands.rs
boomshroom: One of Markdown's strengths is that it's easy to read as a human even without formatting. Auto-renumbering kind of hurts that and I've found more cases where I wish it didn't happen than did. 

In this case, even if it can be auto-renumbered, I'd still say to number them manually so that it's easier to read when looking at the raw source code rather than just the docs. (Especially since this is an executable rather than a library, so the source code would likely be read more than the docs.)

midischwarz12: I really don't see how in any way this is hurting readibility. It isn't like the numbers are being referenced in any sort of way. Which number that the item in the ordered list doesn't really effect how you interpret this at all. It could just as well be an unordered list, but an ordered list was used to communicate precedence.

henriquekirchheck: I think I'll still change it to the actual numbers, this should represent an ordered list of items, and it will be worst for people reading the source code in the future.

src/commands.rs
```diff
@@ -36,13 +38,80 @@ pub enum EnvAction {
...
+                    program.to_string_lossy(),
+                    auto.to_string_lossy()
+                );
+                Ok(auto)
+            }),
+            ElevationStrategy::Force(program) => Ok(program.into()),
+        }
+    }
+
+    /// Gets a path to a previlege elevation program based on what is available in the system.
```
NotAShelf:
```suggestion
    /// Gets a path to a privilege elevation program based on what is available in the system.
```

src/commands.rs
```diff
@@ -221,82 +290,75 @@ impl Command {
...
-                    cmd = cmd.args(&[
-                        "--set-home",
-                        &format!("--preserve-env={}", preserve_vars.join(",")),
-                    ]);
-                }
+        // Use NH_SUDO_ASKPASS program for sudo if present
+        let program_name = elevation_program
+            .file_name()
+            .and_then(|name| name.to_str())
+            .unwrap_or("");
```
NotAShelf:
```suggestion
let program_name = elevation_program
    .file_name()
    .and_then(|name| name.to_str())
    .ok_or_else(|| eyre!("Failed to determine elevation program name"))?;
```

Silent errors are not very helpful for the user. Let's propagate the actual error.

src/commands.rs
```diff
@@ -36,13 +38,80 @@ pub enum EnvAction {
...
 }
 
+/// Define what strategy to use when choosing privilege scalation programs.
+#[allow(dead_code)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ElevationStrategy {
+    Auto,
+    Prefer(PathBuf),
+    Force(&'static str),
+}
```
NotAShelf:
```suggestion
/// Strategy for choosing a privilege elevation program.
/// - `Auto`: try supported programs in fallback order.
/// - `Prefer(PathBuf)`: try the specified program, then fallback.
/// - `Force(&'static str)`: use only the specified program, error if not available.
```

In case myself or another contributor has trouble remembering what the enum is for.

NotAShelf (changes requested):
Seeing I've had the opportunity to nit a littlee bit, this is probably my last review comment. This is now suitable for a merge, which I will go ahead with after my latest comments are addressed. Good work!

P.S. if you could squash your commits that would be great so that I can avoid a squash merge.

henriquekirchheck:
@NotAShelf Everything should be done now, squashed as requested!

PR #99, opened 2024-04-27
added start curly bracket

kapsikkum:
the example didn't work because it was missing the beginning of the curly bracket


kindaLakshay:
i was actually gonna make a pr for the exact same thing. good stuff.

viperML:
Thank you

Issue #79, opened 2024-04-02
Permission denied when trying to update flake.lock 


Yucklys:
I put my flake.nix under `/etc/nixos`, as this is where the `sudo nixos-rebuild switch` default to. However, when I try to update the flake inputs by `nh os switch --update`, it successfully pulls the updates, but failed to update the flake.lock file due to permission denied. I have to manually update the flake inputs before I run this command. I think nh should ask for sudo permission to update the lock, as `/etc/nixos/flake.nix` is a common place that people will store their flake files.
![image](https://github.com/viperML/nh/assets/34306920/490f0134-e26e-41a5-ab56-f857cc478b36)

viperML:
nh doesn't run nix as root, only the activation scripts

Yucklys:
I just want to know if this is the intended behavior, as it requires another command to upgrade. I write a wrapper for the commands and now it work perfectly.

painsucker2:
![Screenshot_20240419_125001](https://github.com/viperML/nh/assets/157894292/6f0c31ce-3d59-4a1c-a7eb-62d8fc7334bc)
Same issue.

painsucker2:
zsssssssss





















s

> nh doesn't run nix as root, only the activation scripts

I'm guessing that's the issue, as flakes are updated by the nix command?
I'm not that experienced in nix, so excuse me if I'm wrong.

viperML:
I know the issue but not the cleanest way to handle it

painsucker2:
So the temporary fix is changing the ownership of flake.nix and flake.lock?

Different-Name:
I am also struggling with this, nh os switch whenever I update my flake inputs, shouldn't nh ask for sudo like it does at other times it is required?

viperML:
No, because I want to keep the evaluation apart from root

Different-Name:
Do you suggest permanently changing ownership then?

viperML:
Running nix flake update as the same user that owns the flake

gepbird:
I have a similar issue: using agenix the quick and dirty way and putting a `builtins.readFile config.age.secrets.foo.path` in the config, I get a permission error: *error: opening file '/run/agenix/foo': Permission denied*

And `sudo -E nh os switch` doesn't work: *Don't run nh os as root. I will call sudo internally as needed*. Whats the reason for this sudo check?

viperML:
Well `readFile secret` is certainly a usecase I want to block, so nh was working as intended.

[...]

PR #101, opened 2024-05-01
handle permissions issues for `--update` arg

NotAShelf:
Tries to open `flake.lock` with both read and write permissions (which should, in theory, be enough to avoid failures due to flake.lock being owned by root.

If it fails (e.g., due to insufficient permissions or the file not existing), it bails with the error message and a short warning.

### DIFF
--- a/src/home.rs
+++ b/src/home.rs
@@ -11,7 +11,7 @@ use crate::*;
 use crate::{
     interface::NHRunnable,
     interface::{FlakeRef, HomeArgs, HomeRebuildArgs, HomeSubcommand},
-    util::{compare_semver, get_nix_version},
+    util::{check_perms, compare_semver, get_nix_version},
 };
 
 #[derive(Error, Debug)]
@@ -66,6 +66,8 @@ impl HomeRebuildArgs {
             let nix_version = get_nix_version().unwrap_or_else(|_| {
                 panic!("Failed to get Nix version. Custom Nix fork?");
             });
+
+            check_perms("flake.lock")?;
 
             // Default interface for updating flake inputs
             let mut update_args = vec!["nix", "flake", "update"];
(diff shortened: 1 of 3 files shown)


NotAShelf:
Might address #79

viperML:
I don't know the purpose of this. Nix already complains with `flake.lock: Permission denied`

NotAShelf:
to handle the error ourselves :)

viperML:
another option is wrapping the update command with `.wrap_err` https://docs.rs/eyre/latest/eyre/trait.WrapErr.html#tymethod.wrap_err

viperML:
But thinking about it, an error in the update command might be caused by other things, so it sounds reasonable to pre-check for the flake.lock permission error. Although I would add a message saying that nh doesn't support updating flakes owned by root

NotAShelf:
Error message is a bit better now.

Side note, I was unable to test this because even when my flake.lock is owned by root, it can still update on my system. So, testing on a sane system would be appreciated.

NotAShelf:
Side note, the file check should probably go into `util.rs` as a function since it's the exact same in both home-manager and nixos command interfaces, but I don't have the the time to get on that.

NotAShelf:
Does nh attempt to read flake.lock before the rebuild? Looks like permission denied error is given before we can reach the check.

viperML:
In home.rs, it is read some lines before to check for the HM output

NotAShelf:
I'm currently testing with `nh os switch` - should home.rs still be referenced in this case?

viperML:
no

NotAShelf:
Then I have no idea why we are not getting the custom error message.

viperML:
it's so over

NotAShelf:
nhover

viperML:
@NotAShelf should I close this PR?

NotAShelf:
If you are uninterested in figuring out why the behaviour occurs, feel free.

PR #137, opened 2024-07-26
feat(os & home): add `--result` flag for `build` subcommands

### LINKED ISSUE #133: Add flag to enable creation of `result` symlinks for build commands
On the one hand, it's really nice that `nh * build` doesn't create `result` symlinks, as they (if not cleaned up) prevent garbage collection,
on the other hand it's sometimes really helpful to be able to look around the files that will end up on my system.

Therefore, I would suggest adding a new flag (e.g., `--symlink-result`) to the `nh os build` & `nh home build` commands that creates the `result` symlink.

If you agree that this would be a good feature for `nh` I can also try to implement it 🙂

DerRockWolf:
resolves #133 

I finally had some time to give it a shot :slightly_smiling_face: 
I hope the implementation is ok. First real time writing rust :blush: 

I've successfully verified the behavior for both nixos and home-manager.

Happy to hear you feedback!

### DIFF
--- a/src/home.rs
+++ b/src/home.rs
@@ -1,5 +1,6 @@
 use std::env;
 use std::ops::Deref;
+use std::path::Path;
 use std::path::PathBuf;
 
 use color_eyre::eyre::bail;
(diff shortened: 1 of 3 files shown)


viperML:
This is wrong because the tempfile is dropped, which causes it to be deleted.

viperML:
I used some type-erasure to keep the possibility of either of the types (PathBuf or TempDir) under the same variable, which I liked more.

DerRockWolf:
> I used some type-erasure to keep the possibility of either of the types (PathBuf or TempDir) under the same variable, which I liked more.

No worries, thanks for looking into it :heart:

Issue #88, opened 2024-04-18
`FLAKE` is not a very specific env var


dpc:
I just watched https://www.youtube.com/watch?v=DnA4xNTrrqY and I like the idea.

`FLAKE` is a very generic name for an env var and I'd prefer something like `NIXOS_CONF_FLAKE` or `NH_OS_FLAKE` (good practice to prefix globals like this with a name of a tool).

viperML:
I've had `NH_FLAKE` in the back of my mind, but at the same time haven't found any project that also uses `FLAKE`. Perhaps it is possible to parse both from the environment, with priority to the latter

dpc:
Definitely possible. I thought about `_OS_` if `nh` was to handle more stuff. I don't know if that's the plan. One might have "NixOS configuration flake" and then some other flakes.

Adda0:
I agree that something like `NH_OS_FLAKE` or even `NH_NIXOS_FLAKE`, and maybe in the future `NH_<OTHER>_FLAKE` would be a great descriptive and conflict-free solution.

viperML:
You can use [wrapper-manager](https://github.com/viperML/wrapper-manager) to wrap nh and not pollute the global environment

Adda0:
That is a very interesting project indeed. This approach would solve this issue for me.

x3rAx:
I'd love to see something like `NH_OS_FLAKE` and `NH_HOME_FLAKE` used by the commands `nh os` and `nh home` respectively, with a fallback to a general `NH_FLAKE` (or just `FLAKE` or whatever) when the command specific variable is not set

Or maybe don't use `NH_FLAKE` at all and simply support `FLAKE` just for backwards compatibility, since I don't know of a way to have the os/home config in the same flake other using the "home manager in system config" approach. But I guess in that case, you can't build home manager independently fom your system anyways?

Adda0:
To get on the same page, may I ask what is the conclusion from this discussion which resulted in closing the issue? What is the recommented approach we should use now?

viperML:
nh will use `NH_FLAKE` as of 4.0 for any installable

viperML:
For the record, I don't know how I could reuse the same `Installable` clap `Args`, and use a different env variable depending on the context.

viperML:
Related: #174

PR #142, opened 2024-08-08
Specific Env Vars and Edit Command

Burtannia:
# Description

## Specific Env Vars

1) Introduce two new environment variables / arguments: `NH_OS_FLAKE` and `NH_HOME_FLAKE` to set the flake for the NixOS and Home Manager configs respectively.
2) Remove the `FLAKE` argument since it's no longer needed and, as others have pointed out #88, it's not very specific. Note that a fallback is in place to `FLAKE` if `NH_..._FLAKE` is not set so as to make the change a bit more backwards compatible.
3) Fallback to current directory if neither `NH_..._FLAKE` nor `FLAKE` is set.

## Edit Command

Running `nh <os | home> edit` uses the aforementioned env vars to open the default editor (as specified by `EDITOR`) in the flake directory.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -177,12 +177,12 @@ pub fn edit_with(flakeref: FlakeRef, editor: String) -> Result<()> {
         }
     };
 
-    CommandBuilder::default()
-        .args(&vec![editor, flakedir])
-        .message("Editing flake")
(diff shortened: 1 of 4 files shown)


viperML:
I will implement this in #151

PR #146, opened 2024-08-17
Return with exit code when --ask is not accepted

Jeidnx:
This change exits the program with code 1 when the user rejects changing to the new configuration. The motivation behind this change is to make it easier to wrap this program in scripts.

As i have written in the code, i wanted to propagate the error properly but with color_eyre this doesn't seem to be an option.

I was also a bit confused what `drop(out_dir)` was doing there and couldn't understand the comment, so i changed it to something more clear.

### DIFF
--- a/src/nixos.rs
+++ b/src/nixos.rs
@@ -134,7 +134,7 @@ impl OsRebuildArgs {
             let confirmation = dialoguer::Confirm::new().default(false).interact()?;
 
             if !confirmation {
-                return Ok(());
+                bail!("User rejected the new config");
             }
         }
 
--- a/src/nixos.rs
+++ b/src/nixos.rs
@@ -175,7 +175,8 @@ impl OsRebuildArgs {
                 .exec()?;
         }
 
-        // Drop the out dir *only* when we are finished
+        // Make sure out_path is not accidentally dropped
+        // https://docs.rs/tempfile/3.12.0/tempfile/index.html#early-drop-pitfall
         drop(out_path);
 
         Ok(())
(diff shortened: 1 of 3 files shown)


src/nixos.rs
```diff
@@ -118,7 +118,12 @@ impl OsRebuildArgs {
             let confirmation = dialoguer::Confirm::new().default(false).interact()?;
 
             if !confirmation {
-                return Ok(());
+                // exit doesn't call destructors
+                drop(out_dir);
+                // This seems to be the only way to exit non-zero
+                // without printing more text to stdout.
```
viperML: Why not call `bail!("some message");`? color_eyre prints to stderr

Jeidnx: Oh wow i didn't even see that you already reviewed this last week, Github didn't show that to me somehow.

I didn't mean stdout specifically, my wording there is misleading. What i meant was that bail prints an Error message to the terminal. I don't think its appropriate to print an error here, since there is nothing wrong with the code / program. Afaict there is no way to use / extend eyre in a way that allows returning a Report without printing anything to the Terminal.

viperML: It was my fault, I wrote the review message one week ago but forgot to click on the submit button.

In any case, I think displaying an error message if the exit code != 0 is not that bad.

```
> Apply the config?
no
Error:
   0: User didn't confirm the request

Location:
   src/nixos.rs:137
```

Jeidnx:
I've made the requested change and also split the changes into two commits.

viperML:
Can you apply the same src/home.rs and src/clean.rs? Then it should be ready

Jeidnx:
Of course, done

PR #150, opened 2024-09-03
add profile naming functionality

GuterBenutzername:
Adds the ability to name profiles, a la nixos-rebuild --profile-name

Quick, dirty, but functional; cleanup is appreciated.

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -128,6 +128,10 @@ pub struct CommonRebuildArgs {
     #[arg(long, short)]
     pub ask: bool,
 
+    /// Name of the new system profile in the bootloader
+    #[arg(long, short)]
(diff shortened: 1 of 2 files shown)


GuterBenutzername:
for naming profiles when added to the bootloader, for if you prefer to be able to have readable names on boot instead of "Generation 53", "Generation 54", etc.
nixos-rebuild had this feature, so I just reimplemented it here.

viperML:
Doesn't system.nixos.tags already give you some customizability?

GuterBenutzername:
Yes, but it's useful to have the option, particularly in my usage scenario where it's easier to name profiles after the changes that were made compared to the last one (like a git commit message)

viperML:
It doesn't work with systemd-boot.

Issue #66, opened 2024-02-05
nh {os, home} repl


viperML:

dev-null-undefined:
Shouldn't this be as simple as `nix repl --expr "builtins.getFlake \"$FLAKE\""`

diniamo:
nixos-rebuild has a built-in option as well:
```sh
nixos-rebuild --flake "$FLAKE" repl
```

viperML:
Added in NH 4.0

PR #171, opened 2024-11-14
nh os repl: init

NotAShelf:
Adds a generic implementation of a `nix repl` wrapper, and a repl subcommand for `nh os` which puts the user into a repl with the active nixosConfiguration. Equivalent of `nix repl .#nixosConfiguration.$HOSTNAME` or entering nix repl and running `:lf nixosConfigurations.<hostname>`.

home repl is not implemented because 1. I conceptually disagree with standalone home-manager and 2. it would depend on functions defined in home.rs, which I would rather not pass around haphazardly - might be implemented properly when there is a more standard library/utility module in nh's codebase.

### DIFF
--- a/src/nixos.rs
+++ b/src/nixos.rs
@@ -1,5 +1,5 @@
 use std::ops::Deref;
-use std::{env, vec};
+use std::vec;
 
 use color_eyre::eyre::{bail, Context};
 use color_eyre::Result;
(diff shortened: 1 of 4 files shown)


NotAShelf:
Sorta fixes #66

src/nixos.rs
```diff
@@ -21,13 +21,47 @@ impl NHRunnable for interface::OsArgs {
...
-impl OsRebuildArgs {
-    pub fn rebuild(&self, rebuild_type: &OsRebuildType) -> Result<()> {
+impl CommonReplArgs {
+    pub fn repl(&self) -> Result<()> {
+        let mut repl_command = vec!["nix", "repl"];
+
+        let flakeref = format!(
+            "{}#nixosConfigurations.{}",
+            self.flakeref.as_str(),
+            hostname::get()
```
viperML: This should use self.hostname if it's not None

viperML (changes requested):

viperML (approved):

NotAShelf:
I should probably bump version in Cargo.toml

Issue #182, opened 2024-11-26
Test nix-darwin in CI


viperML:
I need you to set this up @ToyVo

ToyVo:
I will take a look at this this weekend / holiday

Eveeifyeve:
Anything else that we are missing to test or is this issue completed?

ToyVo:
I have no plans for more on this issue

Eveeifyeve:
I feel like the testing could be a bit better I feel like we should use the tests directory to run tests for nix-darwin.

Eveeifyeve:
Ima try and fix all of the issues so we can get 4.0 under away since it's been 5 months.

Eveeifyeve:
Well sadly I have to update the minium rust version for `dev-dependacies`.

faukah:
Has been resolved by #187.

PR #187, opened 2024-12-01
Test switching to darwin configuration

ToyVo:
Solves #182, I've also addressed a bug that I forgot about when I first implemented the darwin functionality involving how sudo behaves on mac.

The macos runners for github actions use minutes at 10x the rate of linux runners, and the version doesn't effect it, so I went for latest.

### DIFF
--- a/src/clean.rs
+++ b/src/clean.rs
@@ -59,10 +59,7 @@ impl interface::CleanMode {
 
                 // Most unix systems start regular users at uid 1000+, but macos is special at 501+
                 // https://en.wikipedia.org/wiki/User_identifier
-                #[cfg(target_os = "linux")]
-                let uid_min = 1000;
-                #[cfg(target_os = "macos")]
-                let uid_min = 501;
+                let uid_min = if cfg!(target_os = "macos") { 501 } else { 1000 };
                 let uid_max = uid_min + 100;
                 debug!("Scanning XDG profiles for users 0, ${uid_min}-${uid_max}");
                 for user in unsafe { uzers::all_users() } {
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -63,7 +63,27 @@ impl Command {
 
     pub fn run(&self) -> Result<()> {
         let cmd = if self.elevate {
-            Exec::cmd("sudo").arg(&self.command).args(&self.args)
+            let cmd = if cfg!(target_os = "macos") {
+                // Check for if sudo has the preserve-env flag
+                Exec::cmd("sudo").args(
+                    if Exec::cmd("sudo")
+                        .args(&["--help"])
+                        .stderr(Redirection::None)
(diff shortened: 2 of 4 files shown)


flake.nix
```diff
@@ -1,10 +1,18 @@
 {
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.05";
+    nix-darwin = {
```
viperML: I don't want this flake to depend on nix-darwin, as it's only used for tests.

src/commands.rs
```diff
@@ -63,7 +63,27 @@ impl Command {
 
     pub fn run(&self) -> Result<()> {
         let cmd = if self.elevate {
-            Exec::cmd("sudo").arg(&self.command).args(&self.args)
+            #[cfg(not(target_os = "macos"))]
+            let cmd = Exec::cmd("sudo");
+
+            // Check for if sudo has the preserve-env flag
+            #[cfg(target_os = "macos")]
+            let cmd = Exec::cmd("sudo").args(
```
viperML: Why is this needed?

viperML (changes requested):

src/clean.rs
```diff
@@ -59,7 +59,7 @@ impl interface::CleanMode {
 
                 // Most unix systems start regular users at uid 1000+, but macos is special at 501+
                 // https://en.wikipedia.org/wiki/User_identifier
-                #[cfg(target_os = "linux")]
+                #[cfg(not(target_os = "macos"))]
```
viperML: As an stylistic change I'd rather write:

```rust
    if cfg!(target_os = "macos") {
        // macos path
    } else {
        // generic path
    }
```

viperML (changes requested):

src/commands.rs
ToyVo: When invoking sudo on macos we would get errors in the console something along the lines of not owning the home directory
https://github.com/ToyVo/nh_plus/issues/5

```
DEBUG nh::commands:56: cmd=Exec { /usr/bin/sudo nix-env --profile /nix/var/nix/profiles/system --set /var/folders/xm/t9x8767d1p73___d71x35j8r0000gn/T/nh-os-mWyem4/result }
warning: $HOME ('/Users/caleb') is not owned by you, falling back to the one defined in the 'passwd' file ('/var/root')
```

Taken from darwin-rebuild, we could just not support 10.13 and just assume the flag is always there. https://github.com/LnL7/nix-darwin/blob/master/pkgs/nix-tools/darwin-rebuild.sh#L24-L34

ToyVo:
@viperML I've fixed up these few items

Parsifa1:
It's weird, it seems that `@result` will still be generated in `pwd` when using `nh darwin switch`

PR #192, opened 2024-12-06
fix: do not generated `result`  in pwd without `-o` args

### LINKED ISSUE #191: `@result` will be generated in pwd when using `nh darwin switch` without use `-output-path`
https://github.com/viperML/nh/pull/187

It's weird, it seems that `@result` will still be generated in pwd when using nh darwin switch

i tested on my nixos, received the same.

```shell
sudo nix build --profile /nix/var/nix/profiles/system /tmp/nh-osqL0PnV/result
```
image

for this command, maybe we can use `--no-link` ?

'cause `@result` are created by the previous command
```shell
nix build '/home/xxx/nixos/#nixosConfigurations.nixos.config.system.build.toplevel' --log-format internal-json --verbose --out-link /tmp/nh-osqL0PnV/result | nom --json
```
image

if it's not wrong, i can make a pr :)

Parsifa1:
fix #191

have tested the `-o` function.

### DIFF
--- a/src/darwin.rs
+++ b/src/darwin.rs
@@ -127,7 +127,7 @@ impl DarwinRebuildArgs {
 
         if let Switch = variant {
             Command::new("nix")
-                .args(["build", "--profile", SYSTEM_PROFILE])
+                .args(["build", "--no-link", "--profile", SYSTEM_PROFILE])
                 .arg(out_path.get_path())
                 .elevate(true)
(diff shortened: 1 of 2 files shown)


viperML:
Ah yes, at some point I changed from `nix-env` to `nix build` to install the profiles, but I missed that it would create the `result` symlink.

PR #208, opened 2025-01-09
Add argument --vm to nh os build to build a VM start script.

cdo256:
This commit adds the boolean argument `--vm/-v` to build a virtual machine activation script instead of a full system. Note that this only applies to the `os build` subcommand, since it doesn't entirely make sense to switch to a VM, as I don't believe it's currently possible to switch between guest versions on the host or the guest directly without restarting the guest.

The patch https://github.com/viperML/nh/pull/74 was created in May 2024, which proposed to create a build-vm subcommand, but was never fully implemented. This is fully working, though I've only tested it on flakes on NixOS. There's two outstanding todo's:
 - Get it working on Darwin (I don't have hardware to test it currently, though).
 - Fix the logic with the commands to ideally only show the `--vm` option for the `nh os build` subcommand.
 
 Please let me know if you'd be willing to merge a completed patch, and I can sort out those todo's. My feeling is after writing this, that it would make more sense to have a full subcommand like pr 74 suggested. That way it's neater to add VM specific options (such as adding a with --with-bootloader) to act like `#config.system.build.vmWithBootloader`.
 
`build-image` is also possible, but looks to be a little more work since it's handled a bit differently for flakes vs plain Nix. See https://github.com/NixOS/nixpkgs/blob/481c7ead0aafb3d6dfa863ae87e3fdd6badc0b28/pkgs/os-specific/linux/nixos-rebuild/nixos-rebuild.sh#L832

**Just let me know if you'd merge this patch if I finish it off, and whether you'd prefer a build-vm sub-command  or `--vm` option on the build command. I'm happy to do either, but `nh os build-vm` would be my vote.**


viperML:
Yeah, as you probably saw, the types don't allow to have custom flags for each subcommand. You add make more spaguetti types like so, but I'm not sure I'm a fan:

```rust
pub enum OsSubcommand {
    Switch(OsRebuildArgs),
    Build(OsBuildArgs),
    // ...
}

struct OsBuildArgs {
    #[command(flatten)]
    common: OsRebuildArgs,

    #[arg(long)
    vm: bool
}
```

> I'm happy to do either, but nh os build-vm would be my vote.

Indeed

Eveeifyeve:
> * Get it working on Darwin (I don't have hardware to test it currently, though).
I have the hardware to test if needed but there is also a github action that you can use to test.

cdo256:
> Yeah, as you probably saw, the types don't allow to have custom flags for each subcommand. You add make more spaguetti types like so, but I'm not sure I'm a fan:
> 
> ...
> 
> > I'm happy to do either, but nh os build-vm would be my vote.
> 
> Indeed

Thanks confirming, I'll switch the subcommand over to build-vm get a new patch pushed up tonight.

Eveeifyeve:
I might add some nix ci testing to test all the functions.

viperML:
I've been busy these days, so I'll take a deeper look and merge ASAP

src/nixos.rs
```diff
@@ -63,13 +76,18 @@ impl OsRebuildArgs {
             update(&self.common.installable, self.update_args.update_input)?;
         }
 
-        let hostname = match &self.hostname {
+        let system_hostname = hostname::get()
+            .context("Failed to get hostname")?
+            .to_str()
+            .unwrap()
```
NotAShelf: I'm not really a huge fan of `unwrap()`. Bit of a nit maybe, but I think we can do better. For example:

```rs
.and_then(|h| h.to_str().map(String::from).ok_or_else(|| anyhow!("Failed to convert hostname to string")))?;
```

src/nixos.rs
```diff
@@ -22,15 +22,16 @@ impl interface::OsArgs {
...
+            OsSubcommand::Test(args) => args.rebuild(Test, None),
+            OsSubcommand::Switch(args) => args.rebuild(Switch, None),
             OsSubcommand::Build(args) => {
                 if args.common.ask || args.common.dry {
                     warn!("`--ask` and `--dry` have no effect for `nh os build`");
                 }
-                args.rebuild(Build)
+                args.rebuild(Build, None)
             }
+            OsSubcommand::BuildVm(args) => args.build(BuildVm),
```
NotAShelf: Would it not be cleaner to put the logic for `BuildVm` in a separate method? This PR opens the door to more robust builds, but I'd like to avoid overloading build each time if possible (which is sort of what's happening in rebuild.)

```rs
impl OsBuildVmArgs {
    pub fn build_vm(self) -> Result<()> {
        let final_attr = get_final_attr(true, self.with_bootloader);
        self.common.rebuild(OsRebuildVariant::BuildVm, Some(final_attr))
    }
}
```

and then

```diff
-    OsSubcommand::BuildVm(args) => args.build(BuildVm),
+    OsSubcommand::BuildVm(args) => args.build_vm(),
```

src/nixos.rs
```diff
@@ -173,11 +204,26 @@ impl OsRebuildArgs {
...
+    installable: Installable,
+    final_attr: String,
+) -> Installable {
     let mut res = installable.clone();
     let hostname = hostname.as_ref().to_owned();
 
-    let toplevel = ["config", "system", "build", "toplevel"]
+    let toplevel = ["config", "system", "build", &final_attr]
         .into_iter()
         .map(String::from);
```
```diff
-    final_attr: String,
+    final_attr: &str,
 ) -> Installable {
     let mut res = installable.clone();
     let hostname = hostname.as_ref().to_owned();

-    let toplevel = ["config", "system", "build", &final_attr]
+    let toplevel = ["config", "system", "build", final_attr]
         .into_iter()
         .map(String::from);
```

Passing `final_attr` as a string slice (`&str`) ensures it is immutable and avoids unnecessary allocations.

NotAShelf (changes requested):

NotAShelf:
@cdo256 are you still interested in finishing this PR? I've left some comments, after which this could be merged. There are some merge conflicts that arose, but I don't think they are that unmanagable. If you don't plan to finish it, and if you don't mind me taking over, I can try to complete this to help with our #254 goals

cdo256:
@NotAShelf Hi, sorry for the delay. Thanks for the feedback. I'm happy to finish this, I'll see what I can do tonight.

cdo256:
Okay I've think I've covered all of your points. Let me know if there's anything else you think should be changed. I've squashed all of the commits into a single commit, since I'm assuming the intention is for both of them to be merged. (adding build-vm as well as checking if the target and system hostnames match before running `nvd`.) I can split them back up if prefered.

NotAShelf:
The build "fail" appears to be from formatting (which I should fix, I think) and it can be resolved easily. If you could run `sh fix.sh` or invoke `cargo fmt manually` and push your changes again, it should go away. I'll provide a review shortly.

NotAShelf (commented):
LGTM. If I'm not mistaken `final_attr.unwrap_or(String::from("toplevel"))` usage in `toplevel_for` *could* lead to runtime panics *if* an invalid `Option` is provided but seems like the current implementation does ensure that a valid fallback ("toplevel") is always used. 

If you could format the code and add an entry to the changelog in CHANGELOG.md, I can merge this. Some tests could go a long way as well, but I'm currently refactoring the entire testing infrastructure and I won't nag you with them. Thank you for your efforts.

cdo256:
Okay that should have fixed everything, including the most recent merge conflicts. LMK if there's anything else.

NotAShelf (approved):

Issue #174, opened 2024-11-19
Dependant default env variables


viperML:
All nh subcommands use the same `NH_FLAKE`, `NH_ATTRS`, etc env variable.

It would be better if you could also "namespace" this, for example `NH_OS_FLAKE` taking precedence over `NH_FLAKE` for `os` subcommands.

The problem to fix is using different sources for os, home or darwin subcommands.

viperML:
My first idea is to apply the "default values and env variables" at the run site, instead at the clap parsing site.

viperML:
I've implemented something in https://github.com/viperML/nh/pull/179 but I'm not sure it's the best idea.

viperML:
Another option would be to have an specific `Installable` variant for each subcommand, such that it includes the proper description `OsInstallable`, etc, and then map it into `Installable` according to all the fallback values

Eveeifyeve:
So for example darwin switch should be NH_DARWIN_SWITCH_EXTRA_ARGS

viperML:
I think will involve moving parts from the serive to the builder API of clap.

Zynh0722:
I'm not sure this will help you at all, and you probably don't want to bring on another crate for this. 

But for something I'm working on I use the [figment](https://docs.rs/figment/latest/figment) crate, it's can work alongside clap by serializing the clap args as normal and using that as a provider for whatever your main figment is. https://docs.rs/figment/latest/figment/#for-cli-application-authors 

you'd also gain trivial support for config files if that's something you are interested in.

If it doesn't work for you then it doesn't, but I wasn't sure if you had seen figment before.

viperML:
> You want to read configuration from sources outside of the CLI.
> You want flexibility in how configuration sources are combined.
> You want great error messages irrespective of how the application is configured.

It doesn't start well if none of these apply. And I don't see how it would be better to use than the clap builder api.

NotAShelf:
Partially addressed by #250, I'm sure more variables can be added in the future if needed.

PR #212, opened 2025-01-13
Support passing extra args with NH_EXTRA_ARGS environment variable

albinvass:
This adds an environment variable `NH_EXTRA_ARGS` that can be used to pass arguments to nix build as an alternative to positional arguments.

For example these two commands now have the same behaviour:
```
$ nh os switch -- --override-input nixosconfig ~/path/to/local/nixos-flake --no-write-flake-lock

$ export NH_EXTRA_ARGS="--override-input nixosconfig ~/path/to/local/nixos-flake --no-write-flake-lock"
$ nh os switch
```

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -117,13 +117,14 @@ pub struct OsRebuildArgs {
     pub no_specialisation: bool,
 
     /// Extra arguments passed to nix build
-    #[arg(last = true)]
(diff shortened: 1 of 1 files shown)


viperML:
This should also depend on the command, e.g. `NH_OS_EXTRA_ARGS`. I want to implement something for https://github.com/viperML/nh/issues/174 before this

viperML:
Closing in favour of finding a solution for #174 that would integrate this. Thank you for pointing this out in any case.

PR #215, opened 2025-01-15
Clarify that `--ask` is meaningless with `nh ... build` or `nh {home,os} ... --dry`.

ilyagr:
This could be considered a bug report instead of a PR.

I was confused for a while why nh was ignoring my `--ask`, but then realized I was doing `nh home build` instead of `nh home switch`.


viperML:
It should be `warn!` for both cases, not bail or info

ilyagr:
Sure, done.

PR #235, opened 2025-03-09
fix: only warn if FLAKE is set and NH_FLAKE is not

isabelroses:
I set `FLAKE` for other softwares, and currently `nh` will always warn
if `FLAKE` is set dispite me also correctly setting `NH_FLAKE`.


NotAShelf:
Superseded by #250

PR #238, opened 2025-03-18
darwin.rs: handle new root only activation

khaneliman:
Closes https://github.com/viperML/nh/issues/233

Basically, with the new nix-darwin rewrite everything is moved to root activation. I check if the nix-darwin user activation script contains the deprecation notice. ~If it does, we just call the activate script only. (Where the old user activation stuff has been moved)~ Also, moved to using the `darwin-rebuild` script instead of the activation scripts directly so that `nix-darwin` can handle migrations themselves and we don't have to worry about the current implementation details every time a change is made upstream. We just elevate the `activate` when it needs to be. 

Marking draft until the deprecation marker is solidified in https://github.com/LnL7/nix-darwin/pull/1341

### DIFF
--- a/src/darwin.rs
+++ b/src/darwin.rs
@@ -126,36 +126,27 @@ impl DarwinRebuildArgs {
                 .args(["build", "--no-link", "--profile", SYSTEM_PROFILE])
                 .arg(out_path.get_path())
                 .elevate(true)
-                .dry(self.common.dry);
+                .dry(self.common.dry)
+                .run()?;
 
+            let darwin_rebuild = out_path.get_path().join("sw/bin/darwin-rebuild");
             let activate_user = out_path.get_path().join("activate-user");
 
-            let user_activation = Command::new(activate_user.clone())
-                .message("Activating configuration for user")
-                .dry(self.common.dry);
-
-            let activate = out_path.get_path().join("activate");
-
-            let activation = Command::new(activate)
-                .elevate(true)
-                .message("Activating configuration")
-                .dry(self.common.dry);
-
-            if activate_user.exists() {
-                // Check whether activate-user is deprecated
-                // If it is, only activate with root
-                if std::fs::read_to_string(&activate_user)
+            // Determine if we need to elevate privileges
+            let needs_elevation = !activate_user
+                .try_exists()
+                .context("Failed to check if activate-user file exists")?
+                || std::fs::read_to_string(&activate_user)
                     .context("Failed to read activate-user file")?
-                    .contains("# nix-darwin: deprecated")
-                {
-                    activation.run()?;
-                } else {
-                    user_activation.run()?;
-                    activation.run()?;
-                }
-            } else {
-                activation.run()?;
-            }
+                    .contains("# nix-darwin: deprecated");
+
+            // Create and run the activation command with or without elevation
+            Command::new(darwin_rebuild)
+                .arg("activate")
+                .message("Activating configuration")
+                .elevate(needs_elevation)
+                .dry(self.common.dry)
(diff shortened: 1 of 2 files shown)


khaneliman:
Currently calls activation scripts directly, sounds like we should be using the `darwin-rebuild` binary instead.

src/darwin.rs
```diff
@@ -138,18 +138,22 @@ impl DarwinRebuildArgs {
...
-            Command::new(switch_to_configuration)
-                .message("Activating configuration for user")
-                .dry(self.common.dry)
-                .run()?;
-
-            let switch_to_configuration = out_path.get_path().join("activate");
-
-            Command::new(switch_to_configuration)
-                .elevate(true)
+            let darwin_rebuild = out_path.get_path().join("sw/bin/darwin-rebuild");
```
isabelroses, line 132: this introduces a dependency on `darwin-rebuild` which is not always present.

> If you can’t or don’t want to use `darwin-rebuild activate`, then you
  should skip running `activate-user` for post‐user‐activation
  configurations and continue running `activate` as `root`.

khaneliman: In what situation is it not present? This is the `darwin-rebuild` from the built configuration.

khaneliman: This change was done from conversations with emilazy on matrix recommending i switch the implementation from using the activation scripts directly to the built `darwin-rebuild` binary.

isabelroses: > In what situation is it not present? This is the `darwin-rebuild` from the built configuration.

`system.tools.darwin-rebuild.enable = false`

emilazy: You shouldn’t set `system.tools.darwin-rebuild.enable` on a system you want to switch configurations on. We should probably add a stronger warning to that option and maybe hide it, since it’s extremely niche. In the long run we will hopefully have separate `apply`/`switch-to-configuration` scripts, but currently I don’t think this is a configuration activation tools should worry about supporting.

viperML:
Should we check whether the user has an older nix-darwin?

khaneliman:
> Should we check whether the user has an older nix-darwin?

That’s what the deprecation check is for, we see if it’s missing from the activation script. If it is, it’s an older version

khaneliman:
Seems like this will be the way it works in https://github.com/nix-darwin/nix-darwin/pull/1341 . I've been using this branch on my nixos desktop and macbook since I created the PR.

NotAShelf:
The diff looks good. Not large enough for me to nit either, so I think we can merge today.

Though I'm compelled to ask, is this a version specific thing? NH 4.0 has made it into nixpkgs, and changing it now might cause issues. Would that be the case, or can we merge this, release 4.1.0 and PR nixpkgs to provide the new change?

khaneliman:
> The diff looks good. Not large enough for me to nit either, so I think we can merge today.
> 
> Though I'm compelled to ask, is this a version specific thing? NH 4.0 has made it into nixpkgs, and changing it now might cause issues. Would that be the case, or can we merge this, release 4.1.0 and PR nixpkgs to provide the new change?

It's backwards compatible with current nix-darwin. We check for the existence of the deprecation marker or a missing `activate_user` script before using the new method of elevated activation.

NotAShelf:
In that case I can merge this with the addition of a changelog entry in CHANGELOG.md

khaneliman:
Rebased and added changelog entries for changes.

NotAShelf:
Thank you, merging once the CI passes.

khaneliman:
> Thank you, merging once the CI passes.

Thanks raf, appreciate it

PR #239, opened 2025-03-23
search.rs: include platforms in search results

khaneliman:
A common use case, for me, is to find what platforms are supported by a package in nixpkgs. This returns those results in the search command. Not sure if you would rather this be gated behind a flag. 

image

image


khaneliman:
Would you prefer that instead of the flag I added?

PR #244, opened 2025-04-01
fix: get hostname consistently

### LINKED ISSUE #242: `nh home` on darwin should use same hostname logic as `nh darwin`
Here is what we us on `nh darwin`
https://github.com/nix-community/nh/blob/7bd0d00a3f1aab85a7e29d34b4f1f6744263e6c3/src/darwin.rs#L36-L65

`nh home` use the same logic as `nh os`
https://github.com/nix-community/nh/blob/7bd0d00a3f1aab85a7e29d34b4f1f6744263e6c3/src/home.rs#L173-L177

https://github.com/nix-community/nh/blob/7bd0d00a3f1aab85a7e29d34b4f1f6744263e6c3/src/nixos.rs#L66-L73

> I'm using the master branch (`4.0.0-beta.11`), and `nh home switch` doesn't work on darwin, since my flake exports the hostname without `.local`, but `nh` is trying to switch using `.local`.
> 
> The fix in my case would be if `nh` trimmed off the domain part of the hostname with:
> ```rust
> let hostname = hostname::get();
> let hostname.split_once('.').map(|(hostname, _)| hostname).unwrap_or(hostname);
> ```
> For now, I don't know if there's any solution here? I just can't use `nh home` on my darwin setup. 

 _Originally posted by @tqwewe in [#149](https://github.com/nix-community/nh/issues/149#issuecomment-2765995406)_

thecaralice:
Fixes #242

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -74,3 +74,31 @@ impl MaybeTempPath for (PathBuf, TempDir) {
         self.0.as_ref()
     }
 }
+
+pub fn get_hostname() -> Result<String> {
+    #[cfg(not(target_os = "macos"))]
+    {
+        use color_eyre::eyre::Context;
+        Ok(hostname::get()
+            .context("Failed to get hostname")?
+            .to_str()
+            .unwrap()
+            .to_string())
+    }
+    #[cfg(target_os = "macos")]
+    {
+        use color_eyre::eyre::bail;
+        use system_configuration::{
+            core_foundation::{base::TCFType, string::CFString},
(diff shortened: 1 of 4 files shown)


src/util.rs
```diff
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str;
 
-use color_eyre::{eyre, Result};
+use color_eyre::{eyre, eyre::bail, eyre::Context, Result};
```
ToyVo: This is what I get running `fix.sh` in your branch
```suggestion
use color_eyre::{eyre, eyre::bail, Result};
```

PR #248, opened 2025-04-22
generations: clean up; make the implementation faster

NotAShelf:
I've noticed that the `nh os info` command can get a little slow on older hardware. This MR applies some micro-optimizations that I have not bothered with in my PR prior to help with performance on lower-end devices. It yields around 10% faster command execution on my i3-3300Y, and saves around 30ms on my desktop system.


NotAShelf:
Merged on green

PR #251, opened 2025-04-23
build: bump deps

NotAShelf:
Update to NixOS 24.11 (which is the current stable) from NixOS 24.05. Our cargo version is already outdated, and it errors when dependencies are updated.


Eveeifyeve:
prepare for the missile 25.05 is coming out soon.

PR #252, opened 2025-04-24
search: allow caching results; new interface & local endpoints

NotAShelf:
Includes three new changes: channel deprecation, result caching, and fallback search methods. Additionally, those changes are documented in a newly added changelog.

## Changes

We now mark NixOS 24.05 as deprecated, and hard-error if the user attempts to search with a deprecated channnel; NixOS considers those insecure, and so should we.

More importantly, search results can now be cached locally to help with performance of subsequent searches, and reduce network requests. With long cache durations, it might even be possible to support *fully* offline search but I have different ideas on how to approach this. The cache system stores results in the `$HOME/.cache/nh` directory with filenames based on a hash of the query and channel. Cache entries include timestamps and expire after 3600 seconds by default, though this duration is configurable via the new `--cache-duration` option. Users can toggle caching with the `--use-cache` flag. Worth nothing that the cache files are *not* cleaned automatically, maybe the nh module can handle that.

When primary elasticsearch searches fail (happens more often than I'd like to see), the system can now fall back to alternative methods. These include checking a local file specified by `--fallback-file` or querying custom endpoints provided through `--fallback-endpoints`. Fallback endpoints support template variables like `{channel}` and `{query}` for dynamic requests. This mechanism is activated only if the primary search fails and can be disabled with `--use-fallback=false` if so desired.

I've also went ahead and added corresponding environment variables for persistent configuration in the light of Viper's vision.

### Local Search

I made it possible to search locally if you want to avoid network searches. For consistency, the file format should be the same as the Elasticsearch response format

```json
[
  {
    "package_attr_name": "packageName",
    "package_attr_set": "nixpkgs",
    "package_pname": "package-name",
    "package_pversion": "1.2.3",
    "package_platforms": ["x86_64-linux", "aarch64-linux"],
    "package_outputs": ["out", "dev"],
    "package_default_output": "out",
    "package_programs": ["program1", "program2"],
    "package_license_set": ["MIT"],
    "package_description": "A short description.",
    "package_longDescription": "A longer description that might contain more details.",
    "package_hydra": null,
    "package_system": "x86_64-linux",
    "package_homepage": ["https://example.com"],
    "package_position": "pkgs/tools/misc/packageName/default.nix:123"
  },
  {
    "package_attr_name": "anotherPackage",
    "package_attr_set": "nixpkgs",
    "package_pname": "another-package",
    "package_pversion": "0.1.0",
    "package_platforms": ["x86_64-darwin"],
    "package_outputs": ["out"],
    "package_default_output": null,
    "package_programs": [],
    "package_license_set": ["GPLv3"],
    "package_description": null,
    "package_longDescription": null,
    "package_hydra": null,
    "package_system": "x86_64-darwin",
    "package_homepage": [],
    "package_position": null
  }
]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,30 @@
+# NH Changelog
+
+## Unreleased
+
+### Added
+
+- Implement local caching for search results to improve performance on
+  subsequent searches and reduce network requests. Results are stored in
+  `$HOME/.cache/nh`, or in `/tmp/.cache/nh` if `$HOME` cannot be resolved.
+  - add `--use-cache` flag to enable/disable caching (default: enabled).
+  - add `--cache-duration` flag to configure cache expiration time (default:
+    3600 seconds).
+- Introduce fallback mechanisms for situations where the primary search endpoint
+  fails.
+  - add `--use-fallback` flag to enable/disable fallback search (default:
+    `true`).
+  - add `--fallback-file` flag to specify a local file (in Elasticsearch
+    response format) as a fallback data source.
+  - add `--fallback-endpoints` flag to specify custom http endpoints for
+    fallback search, supporting `{query}` and `{channel}` template variables.
+    This is useful if you are using your own search mirror.
+- Enable searching entirely from a local file using the `--fallback-file` option
+  when network searches are undesirable. User is responsible for supplying this
+  file.
+- Add environment variable counterparts for all new command-line flags.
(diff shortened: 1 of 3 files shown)


NotAShelf:
I'm closing this, because the addition of the cache seems to make searches slower overall. Unless the user does repeated searches, this *does* make the cache useless.

PR #259, opened 2025-04-26
search: allow returning output in JSON format

NotAShelf:
Passing `--json`, or defining `NH_SEARCH_JSON`, will now output search results in the JSON format.


viperML (approved):

NotAShelf:
I think the env var is enabled in headless contexts, or just for having it *always* omit JSON. Doesn't really cost us anything, so I'm in favor of having it.

PR #261, opened 2025-04-27
use SUDO_ASKPASS program if set

dvishal485:
A GUI based popup seems more "intuitive" when calling `sudo` internally, as the application just asking for password with sudo on terminal "feels" like hijacking the program. Ofcourse that is not the case, it is completely personal opinion, but you may want to consider this.

Also for those with `sudo` enabled with fingerprint access, it is much more difficult to deal with terminal, as default configuration makes you go for "three fingerprint attempts" before it asks you for password. So if you do not want to use fingerprint, but password instead, normally for programs running with `sudo` (say `sudo echo hello`), you can just <kbd>Ctrl</kbd>+<kbd>C</kbd> during the first attempt, and then password input appears. But since `nh` calls it internally, <kbd>Ctrl</kbd>+<kbd>C</kbd> just halts the execution instead of `nh` instead of letting `sudo` handle it.

Hence running `sudo` with `-A` flag makes sense. But since not everyone have a askpass setup right away, one can conditionally check for the environment variable `SUDO_ASKPASS` (at runtime), and if set, then run `sudo` with `-A` flag to let the ask pass program handle password entry stuff in more "elegant" way.

It will be convenient to both `SUDO_ASKPASS` users and non-users since it will be checked to trigger `sudo`, and hence experience will either remain the same or improve.

Here is an example configuration for nixos for `SUDO_ASKPASS` config which presents a GUI prompt, and if GUI is not available (say running from tty) password is input fallback is done on terminal.

```nix
{ config, pkgs, ... }:
{
  environment.sessionVariables =
    let
      prompt = "Input password for elevated privilages";
      gui-askpass =
        if config.programs.ssh.enableAskPassword then
          "${config.programs.ssh.askPassword} '${prompt}'"
        else
          "${pkgs.zenity}/bin/zenity --password --title='${prompt}'";
      wrapped-askpass = pkgs.writeScriptBin "sudo-askpass" ''
        #!/usr/bin/env sh
        ${gui-askpass} || (read -s -p 'Input Password: ' password && echo $password && unset password)
      '';
    in
    {
      SUDO_ASKPASS = "${wrapped-askpass}/bin/sudo-askpass";
    };
}
```

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -84,8 +84,9 @@ impl Command {
             };
 
             // use NH_SUDO_ASKPASS program for sudo if present
-            let cmd = if std::env::var("NH_SUDO_ASKPASS").is_ok() {
-                cmd.arg("-A")
+            let askpass = std::env::var("NH_SUDO_ASKPASS");
+            let cmd = if let Ok(askpass) = askpass {
+                cmd.env("SUDO_ASKPASS", askpass).arg("-A")
             } else {
                 cmd
             };
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -83,8 +83,8 @@ impl Command {
                 Exec::cmd("sudo")
             };
 
-            // use SUDO_ASKPASS program for sudo if present
-            let cmd = if std::env::var("SUDO_ASKPASS").is_ok() {
+            // use NH_SUDO_ASKPASS program for sudo if present
+            let cmd = if std::env::var("NH_SUDO_ASKPASS").is_ok() {
                 cmd.arg("-A")
             } else {
                 cmd
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -83,6 +83,13 @@ impl Command {
                 Exec::cmd("sudo")
             };
 
+            // use SUDO_ASKPASS program for sudo if present
+            let cmd = if std::env::var("SUDO_ASKPASS").is_ok() {
+                cmd.arg("-A")
+            } else {
+                cmd
+            };
+
             cmd.arg(&self.command).args(&self.args)
         } else {
             Exec::cmd(&self.command).args(&self.args)
--- a/src/main.rs
+++ b/src/main.rs
@@ -55,8 +55,8 @@ fn self_elevate() -> ! {
     use std::os::unix::process::CommandExt;
 
     let mut cmd = std::process::Command::new("sudo");
-    // use SUDO_ASKPASS program for sudo if present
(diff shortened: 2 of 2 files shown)


NotAShelf:
The diff looks good, no notes on that, but I'm not entirely sure if I agree with the intentions of this change. 

> A GUI based popup seems more "intuitive" when calling sudo internally, as the application just asking for password with sudo on terminal "feels" like hijacking the program. Ofcourse that is not the case, it is completely personal opinion, but you may want to consider this.

nh is a command-line program and it seems only natural to me that you would enter the sudo password *in the terminal*. Though I understand it's annoying for fingerprint users (all 3 of them :trollface:) so I can merge this if @viperML agrees with the spirit of the change.

viperML:
I have no idea what SUDO_ASKPASS but if it works then sure

NotAShelf:
`sudo -A` offloads confirmation to a helper program of your choice, `SUDO_ASKPASS` is used to specify the program it will be offloaded to.

dvishal485:
Actually `SUDO_ASKPASS` is not limited to GUI. It can be any program which somehow manages to get password and print it in stdout, hence in theory even a program with hardcoded password with `echo <password>` can also be used as sudo ask pass program, or maybe some gui or tui (like curses or something) can be involved for user to enter password and it just echoes back the password.

`sudo` with -A flag just calls that program and take its output as input for user password instead of handling the password prompt itself.

If `SUDO_ASKPASS` is not set, `-A` flag will throw error msg, hence the check for the flag is done.

NotAShelf:
Just to be explicit, maybe we could listen to a `NH_` prefixed variable. I understand `SUDO_ASKPASS` is standard, but it.might not be desirable for all users to offload `nh` elevation to the helper.

dvishal485:
Ah, but this would be a problem though, sudo uses `SUDO_ASKPASS` env variable to determine which program to call, so checking `NH_SUDO_ASKPASS` will only make sense if it is a boolean. (or `SUDO_ASKPASS` is reassigned as `NH_SUDO_ASKPASS` for execution of sudo)

dvishal485:
That should do it. `nh` should now set `SUDO_ASKPASS` variable from value in `NH_SUDO_ASKPASS`, allowing `nh` to have its own ask pass program configured.

NotAShelf:
Okay, looks good. I'll be reviewing this when I get to my office.

NotAShelf (approved):

Issue #254, opened 2025-04-24
Feature parity with nixos-rebuild


viperML:
If we strive for mass adoption of nh, feature parity is something to consider

- `nixos-rebuild`
  - [x] `switch`
  - [x] `boot`
  - [x] `test`
  - [x] `build`
  - [ ] `dry-build`
  - [ ] `dry-activate`
  - [ ] `edit`
  - [x] `build-vm` (#208)
  - [x] `build-vm-with-bootloader` (#208)
  - [x] `--upgrade` (flakes)
  - [ ] `--upgrade` (channels)
  - [ ] `--install-bootloader`
  - [x] `--no-build-nix`, `--fast` (only this is supported)
  - [x] `--rollback` https://github.com/nix-community/nh/pull/265
  - [x] `--builders` (via passthrough of options)
  - [ ] `--profile-name`
  - [x] `--build-host`, `--target-host` #276 
    - [ ] `--use-substitutes`, `--use-remote-sudo` (nh as a deployment tool)
  - [x] `--flake`
  - [ ] `-I nixos-config=...`

NotAShelf:
I'll take a look at rollback tonight.

cramt:
having the `--target-host` option would be really awesome, i might look into implementing when i have some time

cdo256:
Note that we already have `--dry` which covers `dry-build`, and from looking at the source of nixos-rebuild, it seems that `dry-activate` does pretty much the same thing as `build`, so I can't really see a use case for it.
`edit` just looks up `NIX_CONFIG` (failing that does a search in your `NIX_PATH` for nixos-config) and opens ./default.nix in that directory. I don't think very many people have a ./default.nix in their NixOS config, so this command seems broken by design, and I vote we don't bother including it.
`build-vm` and `build-vm-with-bootloader` are in [pull 208](https://github.com/nix-community/nh/pull/208), and should be nearly ready to merge.

`--install-bootloader` apparently reinstalls the bootloader and can be achieved by simly setting  `NIXOS_INSTALL_BOOTLOADER=1` before running `switch-to-configuraiton`, so that's an easy one.
 `--no-build-nix`, `--fast` both stop nix being updated and `nixos-rebuild` being run a second time. That's not done in `nh` to my knowledge, so these aren't neccessary.
`--rollback` has been merged.

[...]

PR #258, opened 2025-04-26
nh: move checks to a dedicated module; validate nix setup

NotAShelf:
Move nh checks into a separate module, and remove unused functions from utils.

PR #276, opened 2025-05-05
nixos: add `--target-host` and `--build-host` options

cramt:
part of #254

not sure if the right move here is to invoke `ln -s` over ssh to make sure the out_path is symlinked correctly on the target machine, or if just canonicalize'ing the paths is the right move, i went for canonicalize just because it felt less hacky than invoking `ln -s` over ssh

it currently works for me on my own local server

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -17,6 +17,7 @@ pub struct Command {
     command: OsString,
     args: Vec<OsString>,
     elevate: bool,
+    ssh: Option<String>,
 }
 
 impl Command {
@@ -27,6 +28,7 @@ impl Command {
             command: command.as_ref().to_os_string(),
             args: vec![],
             elevate: false,
+            ssh: None,
         }
     }
 
@@ -37,6 +39,11 @@ impl Command {
 
     pub fn dry(mut self, dry: bool) -> Self {
         self.dry = dry;
+        self
+    }
+
+    pub fn ssh(mut self, ssh: Option<String>) -> Self {
+        self.ssh = ssh;
         self
     }
 
@@ -94,9 +101,16 @@ impl Command {
             cmd.arg(&self.command).args(&self.args)
         } else {
             Exec::cmd(&self.command).args(&self.args)
-        }
-        .stderr(Redirection::None)
-        .stdout(Redirection::None);
+        };
+        let cmd = if let Some(ssh) = &self.ssh {
+            Exec::cmd("ssh")
+                .arg("-T")
+                .arg(ssh)
+                .stdin(cmd.to_cmdline_lossy().as_str())
+        } else {
+            cmd
+        };
+        let cmd = cmd.stderr(Redirection::None).stdout(Redirection::None);
 
         if let Some(m) = &self.message {
             info!("{}", m);
@@ -106,9 +120,9 @@ impl Command {
 
         if !self.dry {
             if let Some(m) = &self.message {
-                cmd.join().wrap_err(m.clone())?;
+                cmd.capture().wrap_err(m.clone())?;
             } else {
-                cmd.join()?;
+                cmd.capture()?;
             }
         }
 
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -8,7 +8,7 @@ use subprocess::{Exec, ExitStatus, Redirection};
 use thiserror::Error;
 use tracing::{debug, info};
 
-use crate::{installable::Installable, util::get_current_system};
+use crate::installable::Installable;
 
 fn ssh_wrap(cmd: Exec, ssh: Option<&str>) -> Exec {
     if let Some(ssh) = ssh {
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -4,19 +4,20 @@ use color_eyre::{
     eyre::{bail, Context},
     Result,
 };
-use subprocess::{Exec, ExitStatus, Pipeline, Redirection};
+use subprocess::{Exec, ExitStatus, Redirection};
 use thiserror::Error;
 use tracing::{debug, info};
 
 use crate::{installable::Installable, util::get_current_system};
 
-// for some reason there isnt a common trait that both Exec and Pipeline implements, so just use
-// the no op : from bash
-fn ssh_wrap(cmd: Exec, ssh: Option<&str>) -> Pipeline {
+fn ssh_wrap(cmd: Exec, ssh: Option<&str>) -> Exec {
     if let Some(ssh) = ssh {
-        Exec::cmd("echo").arg(cmd.to_cmdline_lossy().as_str()) | Exec::cmd("ssh").arg("-T").arg(ssh)
+        Exec::cmd("ssh")
+            .arg("-T")
+            .arg(ssh)
+            .stdin(cmd.to_cmdline_lossy().as_str())
     } else {
-        Exec::cmd(":") | cmd
+        cmd
     }
 }
 
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -4,11 +4,21 @@ use color_eyre::{
     eyre::{bail, Context},
     Result,
 };
-use subprocess::{Exec, ExitStatus, Redirection};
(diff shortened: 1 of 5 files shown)


NotAShelf:
This is pretty neat. Viper and I have been discussing how we would handle remote deployments, but I guess simple was how we should've gone about it.

Not on my PC currently, so cannot provide a complete review. Until then, two things

1. Could.you update the CHANGELOG.md found in the repository root? You can add an `## Unreleased` section and add the changes under.

2. Do you think the implementation could be a *little* more generic so that we can also-maybe-possibly consider supporting something like `--build-host` in the future? Not a dealbreaker, just spitballing.

Other than that I have no notes, I'll review in detail when İ get back to my system.

viperML:
The problem with this is that it requires one ssh connection per command, which can be a pain (related: https://github.com/serokell/deploy-rs/issues/54)

I thought about using https://crates.io/crates/libssh-rs to maintain a single SSH connection, but I don't know how hard it would be to implement.

cramt:
the refactor to maintain an ssh connection will probably be a pain, espeically with stuff like shelling out to `nix copy`, but for now i just wanted to mostly rely on the user having ssh-agent so they only need to type their password once, and then have that password type be in the beginning

cramt:
oh i just realized im not handling the nvd diff correct, and that doing so is gonna be real annoying unless nvd just works over ssh

NotAShelf:
I don't think it does, but @bloxx12 and I are currently working on an alternative diff implementation that I'm hoping to replace nvd with in nh as a smaller, faster implementation. Maybe we could also try taking a shot at diff over ssh?

cramt:
> I don't think it does, but @bloxx12 and I are currently working on an alternative diff implementation that I'm hoping to replace nvd with in nh as a smaller, faster implementation. Maybe we could also try taking a shot at diff over ssh?

well diff over ssh is technically not required, if you had like an equivalent to `nvd list` that outputed something machine readible you can just do that on both machines, and then use the outputs to calculate the diff to display locally.

cramt:
managed to get the ssh-add thingy working :tada:

src/util.rs
```diff
@@ -59,6 +59,47 @@ pub fn get_nix_version() -> Result<String> {
...
+            "--expr",
+            "builtins.currentSystem",
+        ])
+        .output()?;
+    let output_str = str::from_utf8(&output.stdout)?;
+    Ok(output_str.to_string())
+}
+
+/// Prompts the user for ssh key login if needed
+pub fn ensure_ssh_key_login() -> Result<()> {
```
NotAShelf, line 46: I don't want to be too nitpicky, but there are three things I would like to note here:

1. Execution of this function appears a little... fragile to me. It assumes `ssh-add` is available, and will degrade user experience if it's missing. Maybe we could check for `ssh-add` early and explicitly.
2. The `.unwrap()` calls (e.g., `Command::new("ssh-add").spawn()?.wait()?`) could cause panics if errors occur. I'm not religiously opposed to unwraps and similar error handling, but for an operation as sensitive as SSH we would like to be*very* explicit.
3. This is a bit of a stretch and mostly optional. `Stdio::inherit` makes sense here for user interaction but could lead to unexpected behavior if nh is run in a non-interactive environment. This will *likely* not be the case since we don't support headless usage (not explicitly, at least) but implicit behaviour *might* be confusing. Take this comment with a grain of salt, I leave it to your best judgement.

src/util.rs
```diff
@@ -59,6 +59,47 @@ pub fn get_nix_version() -> Result<String> {
...
+            "eval",
+            "--impure",
+            "--raw",
+            "--expr",
+            "builtins.currentSystem",
+        ])
+        .output()?;
+    let output_str = str::from_utf8(&output.stdout)?;
+    Ok(output_str.to_string())
+}
```
NotAShelf: Okay I'm *really* not sure about this. Surely there is a way we can get the current system *purely* without having to evaluate `currentSystem`. Maybe we can evaluate `nixpkgs.hostPlatform` or some modular attribute inside the evaluated configuration? CC @viperML on this for installable context related woes.

src/commands.rs
```diff
@@ -8,7 +8,18 @@ use subprocess::{Exec, ExitStatus, Redirection};
...
 
-use crate::installable::Installable;
+use crate::{installable::Installable, util::get_current_system};
+
+fn ssh_wrap(cmd: Exec, ssh: Option<&str>) -> Exec {
+    if let Some(ssh) = ssh {
+        Exec::cmd("ssh")
+            .arg("-T")
+            .arg(ssh)
+            .stdin(cmd.to_cmdline_lossy().as_str())
```
NotAShelf, line 18: Are we sure that `cmd.to_cmdline_lossy()` doesn't pass `unescaped` inputs to the SSH command? I can't exactly imagine an "attack" vector here, but this seems a little dangerous.

NotAShelf (commented):
I like the spirit of those changes, but I really must defer to Viper here for a second opinion. We're messing with SSH, and I'd rather do everything correctly. So far I have a bunch of issues that are plaguing me. The liberal use of `.unwrap()` is one and I find it to be a fragile. Please consider using `?` or `.expect()` with descriptive error messages. Speaking of, error messages would benefit from making a little more explicit. The edge cases should be rare, but not none. Lastly, and this is a little bit speculative, it’s worth validating all inputs that eventually reach shell commands. We do not want to be opening the door to weird command injections, no matter how unlikely.

Also, I'd appreciate some tests but it might be add them without a good mock testing in place. #258 is supposed to address this... eventually.

src/util.rs
cramt: not nipicky at all, youre right, i should make sure to not unwrap at the call site so that we arent ever dependant on ssh-add

src/commands.rs
cramt: I did check that i did do escaping when implementing it, not as much for security as in the case of it would break on weird input like a `'` or smth, i cant vouch for their implementation of escaping being correct, but they atleast do something

src/util.rs
cramt: actually this shouldnt be needed, i can just use `-` to get the `builders` argument to use the default

NotAShelf:
Sorry, there seems to be a small merge conflict. I couldn't fix it on Github's web UI very well. Would you mind fixing it? Looks like we're using both NH's own `Command` struct *and* `Command` from the standard library. Please use the NH's own where applicable.

cramt:
np, thanks for trying, im currently visiting my in-laws, so cant promise it'll be today, so at the latest it'll be sunday

NotAShelf:
No worries, I'll see what I can do if I can get to my system later. I'm also visiting family, so not sure if I'll have the time today.

cramt:
it would require quite a few additions to the internal NH command struct work for the `ensure_ssh_key_login` function as we need the exit code for the first command and we need to inherit stdio and stdout in the second command, so its probably simplest to just stick to the std lib command

NotAShelf (approved):

NotAShelf:
LGTM. I have a few notes, but I think I've nitted this PR enough-- I can cover those later on my own accord. Is this ready to merge?

cramt:
for my sake its ready to merge

NotAShelf:
Thank you for your efforts :)

khaneliman:
Seems like this broke compare changes diff and introduced a warning on every build saying it's guessing host name and it's building for a VM.

NotAShelf:
I'm doing a little cleanup in the platform commands right now. It will be pushed to a new branch later today, and should avoid the breakage as I *think* I caught the issue and fixed it anyway.

emenel:
> Seems like this broke compare changes diff and introduced a warning on every build saying it's guessing host name and it's building for a VM.

fwiw i'm still seeing this message using the latest version from the repo's flake.

NotAShelf:
Please see if https://github.com/nix-community/nh/pull/281 resolves the issue for you. I'm currently preoccupied with personal projects and can't make too much progress on that PR, but it should already contain the fix for the error reported.

Issue #237, opened 2025-03-12
``nh <cmd> repl`` does not work


imvaskel:
It seems that on Lix version 2.92 (not sure if this occurs before), any of the ``repl`` commands do not work. This is because to use a flake in the repl loadable, you need the experimental feature ``repl-flake``. This requirement is not mentioned anywhere. I think that it should either inform the user that that feature is needed or just pass the experimental feature itself.

viperML:
Indeed. Maybe autoconfiguring the child nix process is debatable, but at least nh should show a warning.

imvaskel:
I'm looking at adding this in, is running `nix config show` and parsing out the JSON a good approach for reading the experimental features?

viperML:
Yeah, I was thinking the same. It shouldn't add much overhead, and there's no need to parse the whole json, just what we need

imvaskel:
So there seems to be an issue with this approach: I am running Lix 2.92 and they seem to have changed the command to return a bit enum of the experimental features, while Lix <2.92 and Nix just return an array of the enabled features.

imvaskel:
After further consideration, I don't think that it is worth it to add a check within the code, maybe some sort of FAQ entry about this is the best approach?

PR #277, opened 2025-05-08
nh: validate nix features & version

NotAShelf:
Isolated from #258. We now check for a minimum Nix version (should be equal to nixVersions.latest from the stable branch) and required Nix features (flakes and nix-command) as a part of our startup checks, which previously only checked for NH_FLAKE or/and FLAKE being set. This also moves the checks out of main and to a dedicated module for future maintainability reasons.

Closes https://github.com/nix-community/nh/issues/237

### DIFF
--- a/src/checks.rs
+++ b/src/checks.rs
@@ -1,0 +1,152 @@
+use std::{cmp::Ordering, env};
+
+use color_eyre::{eyre, Result};
+use semver::Version;
+
+use crate::util;
+
+/// Verifies if the installed Nix version meets requirements
+///
+/// # Returns
+///
+/// * `Result<()>` - Ok if version requirements are met, error otherwise
+pub fn check_nix_version() -> Result<()> {
+    if env::var("NH_NO_CHECKS").is_ok() {
+        return Ok(());
+    }
+
+    let version = util::get_nix_version()?;
+    let is_lix_binary = util::is_lix()?;
+
+    // XXX: Both Nix and Lix follow semantic versioning (semver). Update the
+    // versions below once latest stable for either of those packages change.
+    // TODO: Set up a CI to automatically update those in the future.
+    const MIN_LIX_VERSION: &str = "2.91.1";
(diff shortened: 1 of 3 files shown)


NotAShelf:
@drupol (pinging as the other nh maintainer)

This new change causes `fixupPhase` to fail during `nix build` because the testing sendbox does not have the required features check. Do you think we can set the required variables (or disable the check via `NH_NO_CHECKS`) only during build? `env.NH_NO_CHECKS` would do for suppressing the warning, but I believe that'll disable the functionality *entirely* even after build.

drupol:
Hello,

I don't have a strong opinion on this yet, do what you think is better, and ping me when it's done, I'll have a look.

drupol (approved):

viperML (approved):

andersk:
This does not work with Determinate Nix, for which `nix-command` and `flakes` are not considered experimental (even if explicitly listed in `extra-experimental-features`).

```console
$ nix --version
nix (Determinate Nix 3.6.1) 2.29.0
$ cat ~/.config/nix/nix.conf 
# WARNING: this file is generated from the nix.settings option in
# your Home Manager configuration at $XDG_CONFIG_HOME/nix/nix.conf.
# Do not edit it!
extra-experimental-features = nix-command flakes
extra-substituters = https://nix-community.cachix.org
extra-trusted-public-keys = nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs=
$ nix config show experimental-features

$ nh home switch .
! Missing required Nix features: nix-command, flakes (nh/src/checks.rs:101)
Error: 
   0: Missing required experimental features. Please enable: nix-command, flakes

Location:
   src/checks.rs:105
```

NotAShelf:
We only support mainsteam Nix, and Lix to a degree.

viperML:
It shouldn't be too hard to support Dix for this. In any case we won't try too hard to support Dix if it start to diverge too mucho from Nix for other things...

PR #281, opened 2025-05-11
treewide: consolidate logic in platform-specific commands

NotAShelf:
Welcome back funny PR.

Something something consolidates duplicate platform logic something. Hopefully not a breaking change but I'm pushing this while *incredibly* sick and will go back to bed right after.

### DIFF
--- a/src/util/platform.rs
+++ b/src/util/platform.rs
@@ -10,9 +10,12 @@ use tracing::{debug, info, warn};
 use crate::commands;
 use crate::installable::Installable;
 
-/// Resolves an Installable from an environment variable, or falls back to the provided one.
-pub fn resolve_env_installable(var: &str, fallback: Installable) -> Installable {
-    if let Ok(val) = env::var(var) {
+/// Resolves an Installable from an environment variable.
+///
+/// Returns `Some(Installable)` if the environment variable is set and can be parsed,
+/// or `None` if the environment variable is not set.
+pub fn resolve_env_installable(var: &str) -> Option<Installable> {
+    env::var(var).ok().map(|val| {
         let mut elems = val.splitn(2, '#');
         let reference = elems.next().unwrap().to_owned();
         let attribute = elems
@@ -23,9 +26,7 @@ pub fn resolve_env_installable(var: &str, fallback: Installable) -> Installable 
             reference,
             attribute,
         }
-    } else {
-        fallback
-    }
+    })
 }
 
 /// Extends an Installable with the appropriate attribute path for a platform.
--- a/src/util/platform.rs
+++ b/src/util/platform.rs
@@ -337,7 +337,37 @@ pub fn process_specialisation(
 }
 
 /// Execute common actions for a rebuild operation across platforms
-/// This unifies the core workflow that was previously duplicated across platforms
+///
+/// This function handles the core workflow for building and managing system
+/// configurations across different platforms (`NixOS`, Darwin, Home Manager).
+/// It unifies what would otherwise be duplicated across platform-specific modules.
+///
+/// The function takes care of:
+/// 1. Properly configuring the attribute path based on platform type
+/// 2. Building the configuration
+/// 3. Handling specialisations where applicable
+/// 4. Comparing the new configuration with the current one
+///
+/// # Arguments
+///
+/// * `installable` - The Nix installable representing the configuration
+/// * `config_type` - The configuration type (e.g., "nixosConfigurations", "darwinConfigurations")
+/// * `extra_path` - Additional path elements for the attribute path
+/// * `config_name` - Optional hostname or configuration name
+/// * `out_path` - Output path for the build result
+/// * `extra_args` - Additional arguments to pass to the build command
+/// * `builder` - Optional remote builder to use
+/// * `message` - Message to display during the build process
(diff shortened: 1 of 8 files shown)


NotAShelf:
I have neither standalone HM nor a Darwin configuration in my flake, so I can only *hope* those continue to work as intended. @khaneliman if you could see if this PR resolves the previous warning, that'd be neat. Feel free to also test other Darwin functionality.


For now I'm going back to bed, this can *probably* be merged after proper testing and some rustdoc added to new functions (since they're a part of the util module.)

NotAShelf:
mfw regressed again, god dammit

NotAShelf:
This should be "stable" enough for testing now. As I haven't tested HM and Darwin functionality (knowing there are those who track the master branch) I am __not merging this__ until I can get some test results. If only we had regression testing...

khaneliman:
I don't use a standalone home-manager, but I can test darwin and nixos. Seems to have resolved the warning about guessing hostname and building for a VM. But, it still has no diff output between generations.

NotAShelf:
I've disabled diff in Darwin until I implement the new differ, though I might bring it back before this is merged. Does it not display diffs in current master? I'm off my computer and can't really look into the code rn.

khaneliman:
> I've disabled diff in Darwin until I implement the new differ, though I might bring it back before this is merged. Does it not display diffs in current master? I'm off my computer and can't really look into the code rn.

Nah, it doesn't on master either. I didn't realize it was a conscious choice, thought it was an accident. But, this does resolve the warning.

NotAShelf:
It shouldn't be broken on master, at least I haven't noticed that it was. If you could bisect, I might be able to submit a patch later today or most likely tomorrow. Either way I'll bring it back with the new differ (#282) once I get to it.

khaneliman:
> It shouldn't be broken on master, at least I haven't noticed that it was. If you could bisect, I might be able to submit a patch later today or most likely tomorrow. Either way I'll bring it back with the new differ (#282) once I get to it.

When I bisected it was on that PR I made the comment on, I think 477b751e35628165305d4571dff54e61add3af0d ? . Right now I'm just pinning my flake input to the commit merging the darwin root activation support.

src/util/platform.rs
```diff
@@ -0,0 +1,648 @@
...
+use color_eyre::eyre::bail;
+use color_eyre::eyre::WrapErr;
+use color_eyre::Result;
+use tracing::{debug, info, warn};
+
+use crate::commands;
+use crate::installable::Installable;
+
+/// Resolves an Installable from an environment variable, or falls back to the provided one.
+pub fn resolve_env_installable(var: &str, fallback: Installable) -> Installable {
```
viperML: Bikeshed: this should return an `Option<Installable>`, as the callsites favour doing `platform::resolve_env_installable("NH_OS_FLAKE").unwrap_or(self.installable)`

src/util/platform.rs
```diff
@@ -0,0 +1,648 @@
...
+}
+
+/// Extends an Installable with the appropriate attribute path for a platform.
+///
+/// - `config_type`: e.g. "homeConfigurations", "nixosConfigurations", "darwinConfigurations"
+/// - `extra_path`: e.g. ["config", "home", "activationPackage"]
+/// - `config_name`: Optional configuration name (e.g. username@hostname)
+/// - `push_drv`: Whether to push the drv path (platform-specific)
+/// - `extra_args`: Extra args for nix eval (for config detection)
+pub fn extend_installable_for_platform(
```
viperML, line 39: Maybe could have some unit tests.

src/util/platform.rs
```diff
@@ -0,0 +1,648 @@
...
+/// * `no_specialisation` - Whether to ignore specialisations
+/// * `specialisation` - Optional explicit specialisation to use
+/// * `current_profile` - Path to the current system profile for comparison
+/// * `skip_compare` - Whether to skip comparing the new and current configuration
+///
+/// # Returns
+///
+/// The path to the built configuration, which can be used for activation
+#[allow(clippy::too_many_arguments)]
+pub fn handle_rebuild_workflow(
```
viperML, line 398: I like doing a Builder for commands with many options, but this can go for now.

src/util/platform.rs
NotAShelf: I'll probably rebase the tests PR on top of this one so that we can have unit tests, regression tests *and* property tests

src/util/platform.rs
NotAShelf: What can go? The warning or the function?

cdo256:
Thank you for doing all of this 'manual labor', Raf. I agree something had to be done.

One point that might cause an issue is that currently `platform::get_target_hostname()` does both the job of getting the hostname attr, as well as doing the check to decide whether to run `nvd`.  Issue #299 would require the latter logic to be performed after the derivation is built.  `platform::get_target_hostname()` would have to be split into say `platform::get_target_hostname()` and `platform::should_run_nvd()`.

Then I'd could implement a fix for #299 on this branch.

What do you think? Is it worth the extra complexity?

NotAShelf:
Rebasing will be pain, this will have to come at a later date. Most of the changes were cherry picked anyway.

Issue #202, opened 2024-12-28
nh clean all on darwin 


gigamonster256:
Same behavior as https://github.com/ToyVo/nh_plus/issues/5 during `nh clean all`. I believe

https://github.com/viperML/nh/blob/6a69a145b0c7dbd5616bbded512b8bf8b5d2f8a4/src/main.rs#L44-L50

may need some massaging similar to

https://github.com/viperML/nh/blob/6a69a145b0c7dbd5616bbded512b8bf8b5d2f8a4/src/commands.rs#L68-L84

ToyVo:
Hmm I know I've gotten a commit in that uses the set-home sudo flag on Darwin. https://github.com/viperML/nh/blob/master/src/commands.rs#L66-L81 which fixed it on that original issue

viperML:
Paste the output with `--verbose`

gigamonster256:
`nh clean all -v`
```
TRACE nh::logging:86: Logging OK
DEBUG nh:32: Main {
    verbose: true,
    command: Clean(
        CleanProxy {
            command: All(
                CleanArgs {
                    keep: 1,
                    keep_since: Duration(
                        0ns,
                    ),
                    dry: false,
                    ask: false,
                    nogc: false,
                    nogcroots: false,
                },
            ),
        },
    ),
}
DEBUG nh:33: NH_VERSION=4.0.0-beta.5 NH_REV=Some("6a69a14")
DEBUG nh:49: "sudo" "nh" "clean" "all" "-v"
TRACE nh::logging:86: Logging OK
DEBUG nh:32: Main {
    verbose: true,
    command: Clean(
        CleanProxy {
            command: All(
                CleanArgs {
                    keep: 1,
                    keep_since: Duration(
                        0ns,
                    ),
                    dry: false,
                    ask: false,
                    nogc: false,
                    nogcroots: false,
                },
            ),
        },
    ),
}
DEBUG nh:33: NH_VERSION=4.0.0-beta.5 NH_REV=Some("6a69a14")
DEBUG profiles_in_dir: nh::clean:240: return=["/nix/var/nix/profiles/system"] dir="/nix/var/nix/profiles"
DEBUG profiles_in_dir: nh::clean:240: return=["/nix/var/nix/profiles/per-user/root/profile", "/nix/var/nix/profiles/per-user/root/channels"] dir="/nix/var/nix/profiles/per-user/root"
DEBUG nh::clean:64: Scanning XDG profiles for users 0, $501-$601
[...]

PR #280, opened 2025-05-10
treewide: consolidate logic in platform-specific commands

NotAShelf:
I've noticed a lot of code is duplicated in `nixos.rs`, `home.rs` and `darwin.rs`, which makes it a little difficult to add new functionality to add functionality to all platform-specific command at once.

Still no tests, but testing indicates no regression.

PR #284, opened 2025-05-13
fix: self eval passwd warning for darwin

Eveeifyeve:
Fixes https://github.com/nix-community/nh/issues/202

Tested manually full works!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -49,6 +49,10 @@ fn self_elevate() -> ! {
 
     let mut cmd = std::process::Command::new("sudo");
 
+    if cfg!(target_os = "macos") {
+        cmd.args(["--set-home", "--preserve-env=PATH", "env"]);
+    }
+
     // use NH_SUDO_ASKPASS program for sudo if present
     let askpass = std::env::var("NH_SUDO_ASKPASS");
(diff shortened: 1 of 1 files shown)


Eveeifyeve:
@ToyVo I have done the pr it has been 4 months with no work so here it is.

NotAShelf:
1. We do not use conventional commits, but instead *scoped* commits.  The commit message should be `module: description of changes` instead.
2. Please also update the changelog found in the project root.

I've got no Darwin machine to test with, so I can merge this once we confirm that it resolves the original issue. Unfortunately I won't be making a release with this anytime soon because of another Darwin regression with diffs, but we can include this in 4.1.0 once I complete #280.

viperML:
I thought we used conventional commits, damn

NotAShelf:
not anymore :trollface: 

guess we should write some contributing guidelines after the tests PR is done

Eveeifyeve:
> 1. We do not use conventional commits, but instead _scoped_ commits.  The commit message should be `module: description of changes` instead.

Question what would that be if i only edited the src/main.rs

NotAShelf:
In that case `nh: ...` is fine.

PR #291, opened 2025-05-18
Allow switching between specialisations.

daschw:
Prefer switching to specialisation provided as `--specialisation` argument over current specialisation.

---

Currently, assuming a config like in the README

```nix
{config, pkgs, ...}: {
  specialisation."foo".configuration = {
    environment.etc."specialisation".text = "foo";
    # ..rest of config
  };

  specialisation."bar".configuration = {
    environment.etc."specialisation".text = "bar";
    # ..rest of config
  };
}
```

we have this - in my optinion - unexpected behaviour:

```sh
nh os switch -s foo # switch to foo
nh os switch -s bar # switch back to no-specialisation
```

---

After these changes

```sh
nh os switch -s bar
```

and 

```sh
nh home switch -s bar
```

switch to the bar specialisation regardless of the current specialisation.

### DIFF
--- a/src/home.rs
+++ b/src/home.rs
@@ -99,15 +99,13 @@ impl HomeRebuildArgs {
 
         debug!(?prev_generation);
 
-        let spec_location =
-            PathBuf::from(std::env::var("HOME")?).join(".local/share/home-manager/specialisation");
-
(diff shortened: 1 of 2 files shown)


daschw:
I guess this will be fixed by #281 anyways

PR #292, opened 2025-05-18
fix cargo clippy command in fix.sh

daschw:
Running `./fix.sh` errors with

```sh
+ echo 'Running clippy linter and applying available fixes'
Running clippy linter and applying available fixes
+ cargo clippy --fix --allow-dirty -W clippy::pedantic -W clippy::pedantic -W clippy::correctness -W clippy::suspicious
error: unexpected argument '-W' found
```

This fixes it and removes a duplicate line.


NotAShelf (approved):
oops

PR #296, opened 2025-05-21
treewide: fix nvd diffs and internal cleanup

NotAShelf:
This'll create some merge conflicts, but I'd like to get this fixed before I take an extended break. Fixes nvd diffs and supresses a redundant argument that would occur after #208. Also improves debug logging for references that we need to keep track of. Should help track down issues such as this one easier in the future.

CC @khaneliman for Darwin testing, seems to be working as intended on NixOS- testing for Home Manager would also be great.

### DIFF
--- a/src/nixos.rs
+++ b/src/nixos.rs
@@ -459,7 +459,7 @@ fn find_previous_generation() -> Result<generations::GenerationInfo> {
             if let Some(filename) = path.file_name() {
                 if let Some(name) = filename.to_str() {
                     if name.starts_with("system-") && name.ends_with("-link") {
-                        return generations::describe(&path, &profile_path);
+                        return generations::describe(&path);
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


NotAShelf:
Also CC @viperML because I'd like to mark this as the last PR for 4.1.0, it gets the known regressions out of the way and brings the new Darwin activation stuff into a release, which we'll need to hurry up with.

khaneliman:
Diff is working for me on darwin on this branch

src/nixos.rs
```diff
@@ -253,8 +281,12 @@ impl OsRebuildArgs {
                 .run()?;
         }
 
-        // Make sure out_path is not accidentally dropped
+        // Make sure out_path is not acidentally dropped
```
khaneliman, line 284:
```suggestion
        // Make sure out_path is not accidentally dropped
```

viperML (approved):

NotAShelf:
> I don't know if we are turning the codebase into some spaguetti, but the priority is to fix the diffs.

I'll do some actual cleanup after 4.1.0, just interested in getting 4.1.0 out of the way before sit down for a potentially large refactor. #281 is still a priority, but it'll take me a little while.

PR #303, opened 2025-05-28
meta: switch to Nixpkgs 25.05; bump Cargo edition

NotAShelf:
NixOS 25.05 is out, and we're ready to switch. Since there are no user-facing changes, this will not be making it into a new release. We also update to 2024 edition so that integration with [bloxx12/dix](https://github.com/bloxx12/dix) is possible.

## TODO

- [ ] Update `set_var` usage (it is now unsafe, and will cause compiler errors)
- [ ] Fix other deperecations/warnings implied by 2024 edition

## Notes

This will be merged before #281, and it'll probably cause a lot of merge conflicts. As such I'll probably close 281 and re-open it when I have more free time in my hands.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "nh"
 version = "4.1.0"
-edition = "2021"
+edition = "2024"
 license = "EUPL-1.2"
 repository = "https://github.com/nix-community/nh"
 description = "Yet Another Nix Helper"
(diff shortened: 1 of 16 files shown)


NotAShelf:
This technically compiles, but I'm unsure whether we want to keep the `unsafe` blocks as is. Technically it's safe, but do we *really* want to keep things unsafe? We can probably change how we handle the env...

PR #317, opened 2025-06-04
 Support updating multiple specific inputs at once

cole-h:


src/interface.rs
```diff
@@ -448,6 +448,6 @@ pub struct UpdateArgs {
     pub update: bool,
```
cole-h: I was also tempted to rename this to `update_all` and make `update_all` and `update_input` mutually exclusive, but I didn't want to break users who may have been relying on this. I'll send a separate PR (after this merges) for when you're next wanting to bump the semver minor version :)

NotAShelf: I think we can sneak a version bump (4.2.0) sometime this weekend. There is another hotfix I'm planning to send, so feel free to include it in this PR as a separate commit :)

cole-h: A bit late, but pushed up now!

cole-h:
(took it upon myself to add a CHANGELOG entry as well)

PR #319, opened 2025-06-04
Feature test for flakes instead of checking for enabled features

cole-h:
Similar fix to this one for home-manager, and the reasoning is lifted from that same PR: https://github.com/nix-community/home-manager/pull/6824

> Feature testing flakes / nix-command is more robust over configuration
> sniffing. Ultimately, the deciding factor should be if flakes work --
> not if the config looks like they will / won't work.
> 
> This alternative test both asserts that the `nix` command is enabled,
> and that flakes are enabled, without depending on whether or not
> flakes are emitted as an experimental feature.
> 
> This is both repairing support for Determinate Nix 3, and prepares for
> a potential future where Nix itself considers Flakes stable.


NotAShelf:
Thank you for this, but I'm not exactly convinced this ia the way to go... yet.

Currently working on a PR that allows commands to specify their required features, which indirectly supports DetNix and other Nix vendors in the future. If you think we should *still* test for flakes specifically (which I think is too limited in scope, see the removal of `repl-flake` and other possible feature additions/removals in the future) then we can rebase this, though there will be some conflicts.

cole-h:
While I do somewhat feel that "check for the thing you want to use" is preferable to "check for version strings and specific settings" (in most cases), I think it'd also be fine to not do that here. I look forward to seeing your approach, and would be more than happy to help test it!

NotAShelf:
> While I do somewhat feel that "check for the thing you want to use" is preferable to "check for version strings and specific settings" (in most cases),

I agree, but this is maybe not the cleanest approach because we can't check for everything directly, as you did in this PR. For example:

```
let repl_flake_enabled = std::process::Command::new("nix")
            .current_dir(dir.path().to_path_buf())
            .arg("repl")
            .arg(dir.path())
            .stdin(std::process::Stdio::null())
            .stdout(std::process::Stdio::null())
            .stderr(std::process::Stdio::null())
            .status()?;
```

I think this is an unnecessary bottleneck. The output, of course, will be more accurate (unless the API changes) than checking for the experimental features, but I prefer simply looking at experimental features (which we know that are or are not set) and telling the user that they're missing one instead of running the commands and hoping we can handle the output gracefully in a way that covers a large number of use cases. Not to say disagree, I just think checking for experimental features is less... invasive? Regardless I think there is room for checking for flakes specifically since they're a repeated pain point, maybe we can move this to the new `FeatureRequirements` trait I've added, with respect to `NH_NO_CHECKS` centrally.

NotAShelf:
I've [opened a PR](https://github.com/nix-community/nh/pull/320). Testing or/and feedback welcome.

cole-h:
Closing in favor of #320 :)

Issue #309, opened 2025-06-01
Failed to extract version


DADA30000:
~ 
❯ nh os switch /etc/nixos      
Error: 
   0: Failed to extract version

Location:
   src/util.rs:42

sudo nixos-rebuild switch works just fine

DADA30000:
I'm using nix from git, this is the output from `nix --version` 
`nix (Nix) 2.30pre20250521_76a4d4c2`

NotAShelf:
God dammit I forgot to account for git versions...

See if setting `NH_NO_CHECKS` helps.

DADA30000:
> God dammit I forgot to account for git versions...
> 
> See if setting `NH_NO_CHECKS` helps.

yep that works
```
~ 
❯ NH_NO_CHECKS=1 nh os switch /etc/nixos
> Building NixOS configuration
⏱ 0s^C


~ 
❯ nh os switch /etc/nixos
Error: 
   0: Failed to extract version

Location:
   src/util.rs:42

~ 
❯ NH_NO_CHECKS=1 nh os switch /etc/nixos
> Building NixOS configuration
⏱ 0s^C


~ 
❯ 
```

DADA30000:
also there's 1 strange thing, it works for my friend with the same exact nix version, so that's interesting
I'll ask him if he has that env

NotAShelf:
I'm trying to add an exhaustive normalization function for the Nix version, a patch will be available later today.

PR #320, opened 2025-06-05
checks: improve robustness

### LINKED ISSUE #305: Regression 4.0.3 -> 4.1.0 Flakes required when not using flakes
```
~ 
❯ nh os switch -f '<nixpkgs/nixos>' -- -I nixpkgs="$(nixpkgs-path-with-eval)"
! Missing required Nix features: flakes (nh/src/checks.rs:101)
Error: 
   0: Missing required experimental features. Please enable: flakes

Location:
   src/checks.rs:105
```
I don't use flakes.

### LINKED ISSUE #315: nh checks v2
The current nh environment checks are "sufficient", but they create weird conflicts based on the kind of Nix implementation, version or simply the configuration. I propose an alternative, non-monolithic approach to checking experimental features in `check_nix_features` where all features are checked upfront regardless of what command will actually be executed.

What I have in mind is new trait and structs to represent feature requirements. Something like:

```rust
use std::collections::HashSet;

pub trait FeatureRequirements {
    fn required_features(&self) -> Vec<&'static str>;
    fn check_features(&self) -> Result<()> {
        if env::var("NH_NO_CHECKS").is_ok() {
            return Ok(());
        }
        
        let required = self.required_features();
        if required.is_empty() {
            return Ok(());
        }
        
        let missing = util::get_missing_experimental_features(&required)?;
        if !missing.is_empty() {
            return Err(eyre::eyre!(
                "Missing required experimental features for this command: {}",
                missing.join(", ")
            ));
        }
        Ok(())
    }
}

#[derive(Debug)]
pub struct FlakeFeatures;

#[derive(Debug)] 
pub struct LegacyFeatures;

#[derive(Debug)]
pub struct ReplFeatures {
    pub is_flake: bool,
}

impl FeatureRequirements for FlakeFeatures {
    fn required_features(&self) -> Vec<&'static str> {
        vec!["nix-command", "flakes"]
    }
}

impl FeatureRequirements for LegacyFeatures {
    fn required_features(&self) -> Vec<&'static str> {
        vec!["nix-command"]
    }
}

impl FeatureRequirements for ReplFeatures {
    fn required_features(&self) -> Vec<&'static str> {
        let mut features = vec!["nix-command"];
        
        if self.is_flake {
            features.push("flakes");
            
            // Lix-specific repl-flake feature for older versions
            if let Ok(true) = util::is_lix() {
                if let Ok(version) = util::get_nix_version() {
                    if let Ok(current) = semver::Version::parse(&version) {
                        if let Ok(threshold) = semver::Version::parse("2.93.0") {
                            if current < threshold {
                                features.push("repl-flake");
                            }
                        }
                    }
                }
            }
        }
        
        features
    }
}
```

Then we would extend the command interfaces to include feature requirement information, likely as follows:

```rust
use crate::checks::FeatureRequirements;

impl NHCommand {
    pub fn get_feature_requirements(&self) -> Box<dyn FeatureRequirements> {
        match self {
            Self::Os(args) => args.get_feature_requirements(),
            Self::Home(args) => args.get_feature_requirements(), 
            Self::Darwin(args) => args.get_feature_requirements(),
            Self::Search(_) => Box::new(checks::LegacyFeatures),
            Self::Clean(_) => Box::new(checks::LegacyFeatures),
            Self::Completions(_) => Box::new(checks::LegacyFeatures),
        }
    }

    pub fn run(self) -> Result<()> {
        // Check features specific to this command
        let requirements = self.get_feature_requirements();
        requirements.check_features()?;
        
        match self {
            Self::Os(args) => {
                unsafe {
                    std::env::set_var("NH_CURRENT_COMMAND", "os");
                }
                args.run()
            }
            // ...
        }
    }
}

impl OsArgs {
    fn get_feature_requirements(&self) -> Box<dyn FeatureRequirements> {
        match &self.subcommand {
            OsSubcommand::Repl(args) => {
                let is_flake = matches!(args.installable, Installable::Flake { .. }) ||
                              env::var("NH_OS_FLAKE").is_ok();
                Box::new(checks::ReplFeatures { is_flake })
            }
            _ => {
                // Check if using flakes based on installable or environment
                if self.uses_flakes() {
                    Box::new(checks::FlakeFeatures)
                } else {
                    Box::new(checks::LegacyFeatures)
                }
            }
        }
    }
    
    fn uses_flakes(&self) -> bool {
        // Check environment variables first
        if env::var("NH_OS_FLAKE").is_ok() {
            return true;
        }
        
        // Check installable type based on subcommand
        match &self.subcommand {
            OsSubcommand::Switch(args) | 
            OsSubcommand::Boot(args) |
            OsSubcommand::Test(args) |
            OsSubcommand::Build(args) => {
                matches!(args.common.installable, Installable::Flake { .. })
            }
            OsSubcommand::Repl(args) => {
                matches!(args.installable, Installable::Flake { .. })
            }
            _ => false
        }
    }
}
```

and lastly, the global one-off check would be removed. As such we wouldl only check features actually needed for the specific command being executed and take into account whether flakes are being used based on installable type and environment variables.

NotAShelf:
Changes Nix environment checks to a trait-based, less monolithic design to avoid redundant feature checks for commands that don't require checked features in the first place. There is a performance impact, but in my experience it is... not noticable as per my testing.

Regardless I think this is more robust in design, and introduces further future-proofing the current implementation is not fully capable of. Also provides groundwork for supporting more Nix variants in the future cleanly, without utility functions to check everything one-by-one.

Fixes #305, #309

Closes #315

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,19 @@
-<!-- markdownlint-disable no-duplicate-headings -->
+<!-- markdownlint-disable no-duplicate-heading -->
 
 # NH Changelog
+
+## Unreleased
+
+### Changed
+
+- Nh checks are now more robust in the sense that unnecessary features will not
+  be required when the underlying command does not depend on them.
+
+### Fixed
+
+- Nh will now correctly detect non-semver version strings, such as `x.ygit`.
+  Instead of failing the check, we now try to normalize the string and simply
+  skip the check with a warning.
 
 ## 4.1.1
 
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -216,7 +216,7 @@ pub struct OsRebuildArgs {
 impl OsRebuildArgs {
     pub fn uses_flakes(&self) -> bool {
         // Check environment variables first
-        if env::var("NH_OS_FLAKE").is_ok() {
+        if env::var("NH_OS_FLAKE").is_ok_and(|v| !v.is_empty()) {
             return true;
         }
 
@@ -466,7 +466,7 @@ pub struct HomeRebuildArgs {
 impl HomeRebuildArgs {
     pub fn uses_flakes(&self) -> bool {
         // Check environment variables first
-        if env::var("NH_HOME_FLAKE").is_ok() {
+        if env::var("NH_HOME_FLAKE").is_ok_and(|v| !v.is_empty()) {
             return true;
         }
 
@@ -494,7 +494,7 @@ pub struct HomeReplArgs {
 impl HomeReplArgs {
     pub fn uses_flakes(&self) -> bool {
         // Check environment variables first
-        if env::var("NH_HOME_FLAKE").is_ok() {
+        if env::var("NH_HOME_FLAKE").is_ok_and(|v| !v.is_empty()) {
             return true;
         }
 
@@ -567,7 +567,7 @@ pub struct DarwinRebuildArgs {
 impl DarwinRebuildArgs {
     pub fn uses_flakes(&self) -> bool {
         // Check environment variables first
-        if env::var("NH_DARWIN_FLAKE").is_ok() {
+        if env::var("NH_DARWIN_FLAKE").is_ok_and(|v| !v.is_empty()) {
             return true;
         }
 
@@ -589,7 +589,7 @@ pub struct DarwinReplArgs {
 impl DarwinReplArgs {
     pub fn uses_flakes(&self) -> bool {
         // Check environment variables first
-        if env::var("NH_DARWIN_FLAKE").is_ok() {
+        if env::var("NH_DARWIN_FLAKE").is_ok_and(|v| !v.is_empty()) {
             return true;
         }
 
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::path::PathBuf;
 
 use anstyle::Style;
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,19 @@
-<!-- markdownlint-disable no-duplicate-headings -->
+<!-- markdownlint-disable no-duplicate-heading -->
 
 # NH Changelog
 
+## Unreleased
+
+### Changed
+
+- Nh checks are now more robust in the sense that unnecessary features
```
faukah: Missing words.

NotAShelf: oops

I got distracted and fixed the line below (lol)

src/interface.rs
```diff
@@ -164,6 +213,18 @@ pub struct OsRebuildArgs {
     pub build_host: Option<String>,
 }
 
+impl OsRebuildArgs {
+    pub fn uses_flakes(&self) -> bool {
+        // Check environment variables first
+        if env::var("NH_OS_FLAKE").is_ok() {
```
cole-h: Tiny nit: it would be slightly more robust if you checked for some contents (or at least `!is_empty()`), so that if someone has it set in their system environment for whatever reason, it's as simple as disabling it with `NH_OS_FLAKE=` for the single invocation instead of having to figure out how to unset it with `unset` or `set -e` something.

But I realize that might be a little more churn than you want in this PR, so no worries if you don't want to do it right now (or ever)

NotAShelf: I see what you mean, I'll fix it while I've already touched the part.

NotAShelf: Addressed in 7e863e42e6b9c119d4bb5ec316482b98c34f97cc

NotAShelf:
Actually, I think I want to add some regression tests while I'm at it. Property testing is probably not enough.

Issue #299, opened 2025-05-24
nvd diff output not seen


Arjun31415:
I do not see the nvd diff output after successful build. It stopped working only recently and I have no clue what the issue is. 

Input command - 
```
export NIXPKGS_ALLOW_UNFREE=1 && nix flake update && nh os switch ./ -H omen --ask -- --impure --show-trace --max-jobs 1
```

Output now - 
```
these 7 derivations will be built:
  /nix/store/404hq2g6bv2fvx54a0vx5nikrr0xdypk-dunst.service.drv
  /nix/store/lmwm6klnarlgzlxnyy3jgx04wwfxwqkk-home-manager-files.drv
  /nix/store/8p52ykh0xia70nlpd72l0w6gpzrx7a82-home-manager-generation.drv
  /nix/store/dcc5k6v2iy3jmw1c2nj6vz62a1amhj5j-unit-home-manager-prometheus.service.drv
  /nix/store/jb3300n7qmlp3znvgs2mwgqws4rdplk9-system-units.drv
  /nix/store/1lj70dzf588x9vd3iw7d74w3bj1n5cfh-etc.drv
  /nix/store/rwf2wskhgnak5v2d1pipmkykad1w881k-nixos-system-Omen-25.11.20250520.2795c50.drv
dunst.service> building '/nix/store/404hq2g6bv2fvx54a0vx5nikrr0xdypk-dunst.service.drv'
home-manager-files> building '/nix/store/lmwm6klnarlgzlxnyy3jgx04wwfxwqkk-home-manager-files.drv'
home-manager-generation> building '/nix/store/8p52ykh0xia70nlpd72l0w6gpzrx7a82-home-manager-generation.drv'
unit-home-manager-prometheus.service> building '/nix/store/dcc5k6v2iy3jmw1c2nj6vz62a1amhj5j-unit-home-manager-prometheus.service.drv'
system-units> building '/nix/store/jb3300n7qmlp3znvgs2mwgqws4rdplk9-system-units.drv'
etc> building '/nix/store/1lj70dzf588x9vd3iw7d74w3bj1n5cfh-etc.drv'
nixos-system-Omen-25.11.20250520.2795c> building '/nix/store/rwf2wskhgnak5v2d1pipmkykad1w881k-nixos-system-Omen-25.11.20250520.2795c50.drv'
┏━ Dependency Graph with 2 roots:
┃ ✔ impure-nvidia-version-file 
┃                ┌─ ✔ dunst.service 
[...]

PR #326, opened 2025-06-12
Fix "nvd diff output not seen #299"

cdo256:


src/nixos.rs
```diff
@@ -182,6 +183,25 @@ impl OsRebuildArgs {
...
+                    Err(e) => return Err(e).context("Error getting /etc/hostname in the computed profile."),
+                    Ok(false) => return Err(eyre!("/etc/hostname doesn't exist in the computed profile: {}", hostname_file_path.display())),
+                    Ok(true) => {
+                        String::from(fs::read_to_string(&hostname_file_path)
+                            .context("Unable to read /etc/hostname in the computed profile.")?
+                            .trim())
+                    },
+                }
+            },
+        };
```
```suggestion
let dest_hostname = match target_profile.try_exists()? {
            false => return Err(eyre!("Target profile doesn't exist")),
            true => {
                let hostname_file = target_profile.join("etc/hostname");
                if !hostname_file.try_exists()? {
                    return Err(eyre!(
                        "/etc/hostname missing in profile: {}",
                        hostname_file.display()
                    ));
                }
                fs::read_to_string(&hostname_file)?.trim().to_string()
            }
        };
```

Not a big fan of the nestig. Can be simplified as such.

src/nixos.rs
```diff
@@ -182,6 +183,25 @@ impl OsRebuildArgs {
             Some(spec) => out_path.get_path().join("specialisation").join(spec),
         };
 
+        debug!("exists: {}", target_profile.exists());
```
NotAShelf: Is this necessary? We have a check shortly after anyway.

```
let dest_hostname = match target_profile.try_exists() { 
```

*is* the check. I think those two can be consolidated to reduce calls.

cdo256: Makes sense I'll remove this line.

src/nixos.rs
cdo256: I've used your version.

NotAShelf:
Superseded by #352

Issue #325, opened 2025-06-10
`darwin switch` doesn't fail when Homebrew fails


theutz:
If Nix Darwin's build process fails, the nh wrapper simply fails silently. Running `darwin-rebuid switch` exposes the error, so there is a work-around. But it'd be nice if those errors could be exposed.

mettavi:
If have also noticed that the `nh` wrapper skips some useful homebrew output when running `darwin-rebuild switch`:

```
$ nh darwin switch
Homebrew bundle...
...

$ darwin-rebuild switch
Homebrew bundle...
Using kegworks-app/kegworks
Using homebrew/cask
Using homebrew/core
Using jorgelbg/tap
Using bitwarden-cli
Using pinentry-touchid
Using anki
Using calibre
Using carbon-copy-cloner
Using cirrus
Using customshortcuts
Using daisydisk
Using iterm2
Using kegworks
Using keyclu
Using kid3
...
`brew bundle` complete! 28 Brewfile dependencies now installed.
...
```

NotAShelf:
Closed by #334

PR #334, opened 2025-06-27
darwin: show output on activation

khaneliman:
Show the stdout activation logging so users can actually see whats going on.

Shows `nix-darwin` activation logging and the `home manager` verbose logging, when enabled.

Fixes issue reported https://github.com/nix-community/nh/issues/325#issuecomment-2979569306


NotAShelf:
Almost forgot, could you add a changelog entry? This is worth documenting so that users know where the output is coming from.

khaneliman:
> Almost forgot, could you add a changelog entry? This is worth documenting so that users know where the output is coming from.

Sounds good, updated with a changelog entry.

NotAShelf (approved):

NotAShelf:
Thank you :)

PR #335, opened 2025-06-27
cargo: bump version to match tag

khaneliman:
Fix version to match current latest tag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -881,7 +881,7 @@ dependencies = [
 
 [[package]]
 name = "nh"
-version = "4.1.1"
+version = "4.1.2"
 dependencies = [
  "anstyle",
  "chrono",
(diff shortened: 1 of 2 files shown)


viperML:
I don't know if @NotAShelf did something, but we have a GH workflow that should bump the git tag in response to the cargo.toml version changing, but somehow the tag was bumped before.

I'd just bump the cargo.toml to 4.1.3 and let the workflow re-take from there...

NotAShelf:
4.1.3 would be wrong, because the current HEAD includes changes that are *technically* breaking. Which is why I'd like to release 4.2.0 instead. Currently investigating an issue where `switch-to-configuration` is sometimes missing (race condition?) and I'll release 4.2.0 soon. Merging this for now.

PR #338, opened 2025-06-28
Added system-manager support

r0chd:
This pr adds support for [system-manager](https://github.com/numtide/system-manager) which allows to configure system wide settings on non-nixos linux distributions

Also fixed Command output still showing after calling Command::show_output(false)

### DIFF
--- a/src/checks.rs
+++ b/src/checks.rs
@@ -345,6 +345,25 @@ impl FeatureRequirements for DarwinReplFeatures {
     }
 }
 
+/// Feature requirements for System Manager repl commands
+#[derive(Debug)]
+pub struct SystemManagerReplFeatures;
+
+impl FeatureRequirements for SystemManagerReplFeatures {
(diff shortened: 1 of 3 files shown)


r0chd:
`nh system-manager` subcommand is probably bad UX tho

NotAShelf:
In my experience system-manager has been nothing short of unstable. Frankly I am not too confident that I'll be able to maintain support for something I can't get to work, so I must defer to @viperML on wheter this should be a part of the NH.

In any case, I'd like to get a VM test setup going before we think of merging this.

viperML:
I'd prefer system-manager to have the good CLI upstream, instead of deferring to NH.

Edit: thank you for the interest with the PR regardless 😊

r0chd:
Fully understandable, I'll just keep the fork and rebase it from time to time.

There's still the issue with Command::show_output(false), it didn't really suppress any output until I changed redirection from Redirection::None to subprocess::NullFile

PR #339, opened 2025-06-29
Add support for opendoas

CaptainJawZ:
I saw that there is a closed PR that tried to add support to other privilege elevation programs, not sure as to why it was never merged, so I apologize if this PR is inappropriate because of that, that said, if the PR is welcomed here is what I added:

only support for doas, with an environment variable `NH_USE_DOAS`, if set to 1 or true, it will, well... use doas instead of sudo.

To prevent any breaking changes, I kept all the sudo logic within main.rs intact, just enclosed it on a clause to only add flags (--preserve-env and -A) if sudo is the selected helper.

For command.rs and main.rs I also added checks so doas is not run when on macos, but this is temporary, as my job will later this year provide me with a macbook, I could see about adding support for macos, at the moment I can only test on nixos/liunx.

I did a bunch of tests with and without the flag (so testing sudo and doas) and it seems like nothing broke, I hope this helps.  Of course if any changes are required I will be more than welcomed to implement them.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -88,6 +88,8 @@ impl Command {
 
     pub fn run(&self) -> Result<()> {
         let cmd = if self.elevate {
+            // Detect whether the user wants to use doas
+            let usedoas = std::env::var("NH_USE_DOAS").is_ok();
             let cmd = if cfg!(target_os = "macos") {
                 // Check for if sudo has the preserve-env flag
                 Exec::cmd("sudo").args(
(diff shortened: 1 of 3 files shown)


CaptainJawZ:
> This seems like a duplicate of #92 but restricted to only doas. However, I think #92 would be a better option to merge since it is more generic.

yeah it does, but at the time I opened this PR, #92 was abandoned for over six months I believe, I see that it got activity some weeks ago again, which is good, if so, this PR can be closed.

NotAShelf:
Thank you for your efforts @CaptainJawZ. Seeing #92 is picked back up and has a more comprehensive approach, I'll be going forward with it. Though, let the record show that I appreciate your PR. Unfortunately Viper and I could not come to a decision on this before 92 was updated.

Closing in favor of `#92`.

PR #344, opened 2025-07-01
Clean: add option that display cleanup statistics

waspoza:
Very simple patch that adds option to bring back cleanup statistics.


waspoza:
Oops, looks like i was too late and fix is already in.

viperML:
Thank you, but superseded by #343

PR #349, opened 2025-07-08
Inspect `NH_{OS_{HOSTNAME,SPECIALISATION},HOME_{CONFIGURATION,SPECIALISATION},DARWIN_HOSTNAME}` environmental variables

### LINKED ISSUE #348: Default hostname/configuration and specialisation customization
(I am a new user of `nh`, so apologies if I'm overlooking something here.)

Like many people, I use a single flake to manage several NixOS/Home Manager/nix-darwin configurations. While iterating on my configuration today, I found myself running `nh home switch $FLAKE -c user@host` repeatedly. Running `nh home switch --help` helpfully revealed the `NH_HOME_FLAKE` environmental variable, but since my hostname is automatically assigned and differs from that used in my `homeConfigurations` attribute, I still had to specify `-c user@host` repeatedly, as there doesn't appear to be a way to specify a default configuration. I would love to see something like `NH_{HOME_CONFIGURATION,OS_HOSTNAME,DARWIN_HOSTNAME}` environmental variables to specify default values for cases where the automatic inference cannot be relied on.

While looking at the help pages, `--specialisation` stood out to me as another option where specifying a value through an environmental variable might be analogously convenient.

Aehmlo:
This PR uses `clap`'s `env` attribute to specify fallback sources for specifying home configurations, NixOS/Darwin hostnames, and home/NixOS specialisations for situations where the default heuristics are insufficient.

Closes #348.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@
 - `nh darwin switch` now shows the output from the `darwin-rebuild` activation.
   This allows you to see more details about the activation from `nix-darwin`, as
   well as `Home Manager`.
+- The `NH_OS_HOSTNAME`, `NH_OS_SPECIALISATION`, `NH_HOME_CONFIGURATION`,
+  `NH_HOME_SPECIALISATION`, and `NH_DARWIN_HOSTNAME` environmental variables are
+  now consulted when determining what attribute/specialisation to use.
 
 ### Fixed
 
--- a/src/home.rs
+++ b/src/home.rs
@@ -243,7 +243,10 @@ where
                     })?;
 
                 if check_res.map(|s| s.trim().to_owned()).as_deref() == Some("true") {
-                    debug!("Using explicit configuration from flag: {}", config_name);
+                    debug!(
+                        "Using explicit configuration from flag/environment: {}",
+                        config_name
+                    );
                     attribute.push(config_name);
                     if push_drv {
                         attribute.extend(toplevel.clone());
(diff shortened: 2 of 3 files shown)


midischwarz12 (approved):

src/home.rs
```diff
@@ -243,7 +243,10 @@ where
                     })?;
 
                 if check_res.map(|s| s.trim().to_owned()).as_deref() == Some("true") {
-                    debug!("Using explicit configuration from flag: {}", config_name);
+                    debug!(
+                        "Using explicit configuration from flag/environment: {}",
+                        config_name
+                    );
```
NotAShelf, line 249:
```suggestion
                    debug!(
                        "Using explicit configuration from flag/environment: {config_name}",
                        
                    );
```

If I'm not mistaken clippy will tell you this is the correct way. Could also be applied with `sh fix.sh` in the repository root.

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@
 - `nh darwin switch` now shows the output from the `darwin-rebuild` activation.
   This allows you to see more details about the activation from `nix-darwin`, as
   well as `Home Manager`.
+- The `NH_OS_HOSTNAME`, `NH_OS_SPECIALISATION`, `NH_HOME_CONFIGURATION`,
+  `NH_HOME_SPECIALISATION`, and `NH_DARWIN_HOSTNAME` environmental variables are
+  now consulted when determining what attribute/specialisation to use.
```
NotAShelf, line 21: I think we need to be a bit more explicit than "now consulted". Please try to specify the priority and the motive of this change for the potential readers.

NotAShelf (changes requested):
I'm a little unsure if this is the most elegant solution here, but the change is small enough that I'll let it go. If you could address my  comments, I can merge this.

PR #383, opened 2025-08-10
{nixos,home,darwin}: add `edit` subcommand

midischwarz12:
## Sanity Checking

- Style and consistency
  - [ ] I have added appropriate docunentation to new code
- Correctness
        logic
  - [ ] I have documented the motive for those changes in the PR body or commit
        description.
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/src/darwin.rs
+++ b/src/darwin.rs
@@ -8,7 +8,9 @@ use crate::Result;
 use crate::commands;
 use crate::commands::Command;
 use crate::installable::Installable;
-use crate::interface::{DarwinArgs, DarwinRebuildArgs, DarwinReplArgs, DarwinSubcommand, DiffType};
+use crate::interface::{
+    DarwinArgs, DarwinEditArgs, DarwinRebuildArgs, DarwinReplArgs, DarwinSubcommand, DiffType,
+};
 use crate::nixos::toplevel_for;
 use crate::update::update;
 use crate::util::{get_hostname, print_dix_diff};
@@ -33,6 +35,7 @@ impl DarwinArgs {
                 args.rebuild(&Build)
             }
             DarwinSubcommand::Repl(args) => args.run(),
+            DarwinSubcommand::Edit(args) => args.run(),
         }
     }
 }
@@ -227,6 +230,89 @@ impl DarwinReplArgs {
             .with_required_env()
             .show_output(true)
             .run()?;
+
+        Ok(())
+    }
+}
+
+impl DarwinEditArgs {
+    fn run(self) -> Result<()> {
+        let editor = if let Ok(var) = env::var("EDITOR") {
+            var
+        } else {
+            bail!("Unset environment variable `EDITOR`.")
+        };
+
+        // Use NH_DARWIN_FLAKE if available, otherwise use the provided installable
+        let target_installable = if let Ok(darwin_flake) = env::var("NH_DARWIN_FLAKE") {
+            debug!("Using NH_DARWIN_FLAKE: {}", darwin_flake);
+
+            let mut elems = darwin_flake.splitn(2, '#');
+            let reference = match elems.next() {
+                Some(r) => r.to_owned(),
+                None => return Err(eyre!("NH_DARWIN_FLAKE missing reference part")),
+            };
+            let attribute = elems
+                .next()
+                .map(crate::installable::parse_attribute)
+                .unwrap_or_default();
+
+            Installable::Flake {
+                reference,
+                attribute,
+            }
+        } else {
+            self.installable
+        };
+
+        match target_installable {
+            Installable::File { path, .. } => {
+                let str_path = path
+                    .to_str()
+                    .ok_or_else(|| eyre!("Edit path is not valid UTF-8"))?;
+
+                Command::new(editor)
+                    .arg(str_path)
+                    .with_required_env()
+                    .show_output(true)
+                    .run()?
+            }
+            Installable::Flake { reference, .. } => {
+                match &reference[0..5] {
+                    "path:" => {
+                        // Removes `path:` prefix
+                        let path = &reference[5..];
+
+                        Command::new(editor)
+                            .arg(path)
+                            .with_required_env()
+                            .show_output(true)
+                            .run()?
+                    }
+                    _ => {
+                        Command::new("nix")
+                            .arg("flake")
+                            .arg("prefetch")
+                            .arg("-o")
+                            .arg("result")
+                            .arg(reference)
+                            .with_required_env()
+                            .show_output(true)
+                            .run()?;
+
+                        Command::new(editor)
+                            .arg("result")
+                            .with_required_env()
+                            .show_output(true)
+                            .run()?;
+
+                        std::fs::remove_file("result")?
+                    }
+                };
+            }
+            Installable::Store { .. } => bail!("Nix doesn't support nix store installables."),
+            Installable::Expression { .. } => bail!("Nix doesn't support nix expression."),
+        };
 
         Ok(())
     }
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -121,6 +121,13 @@ impl OsArgs {
                 let is_flake = args.uses_flakes();
                 Box::new(OsReplFeatures { is_flake })
             }
+            OsSubcommand::Edit(args) => {
+                if args.uses_flakes() {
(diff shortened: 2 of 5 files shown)


NotAShelf:
Should support non-flakes as well. When an installable is passed or if `--file` is used we'd like to edit it directly.

Also I think we should bail instead of trying nano when $EDITOR is unset. If user doesn't have EDITOR set, then it's unlikely that they have nano installed imo.

NotAShelf:
Also we don't use conventional commits, please consider rewording to match the title.

midischwarz12:
> Should support non-flakes as well. When an installable is passed or if `--file` is used we'd like to edit it directly.

I intentionally left out support for a `--file` flag because at that point, it brings no advantage to just using their desired editor directly since you're already supplying the path.

Although, I can see the usefulness of wanting to automatically direct the editor to a designated flake location or non flake location such as `/etc/nixos`.

NotAShelf:
@midischwarz12 I've moved this to the 4.2.0 milestone. Once again there is no rush, but if you could address my comments as discussed on Matrix I'd like to merge this before Monday and start thinking about 4.2.0

midischwarz12:
I'll probably have time to get this out tomorrow as well. If not, it'll be done on Thursday for sure.

midischwarz12:
@NotAShelf Feel free to take a look. Sorry for the long wait.

src/nixos.rs
```diff
@@ -720,6 +722,72 @@ impl OsReplArgs {
...
+                    .show_output(true)
+                    .run()?;
+
+                Command::new(editor)
+                    .arg("result/flake.nix")
+                    .with_required_env()
+                    .show_output(true)
+                    .run()?;
+
+                std::fs::remove_file("result")?
```
```suggestion
                std::fs::remove_dir_all("result")?;
```

`nix flake prefetch -o result` creates a directory, not a file. Using `remove_file` will always fail and leave junk directories.

NotAShelf (changes requested):
Okay, this PR is less close to a merge than I had hoped. While I *hate* to provide negative reviews, there is no way I can merge this as is. Unfortunately I must leave my desktop shortly, but here is a quick list of thoughts:

1. The duplication of `edit` logic is unnecessary. They are so similar, there could be a generic helper that consolidates all three platform implementations and avoids the absurd +319 LoC addition. You can consider putting this in util, or in commands (not commands.rs; something like `commands::edit`)
2. I hope you do realise there is little to no value in editing a flake from the store directory. Not only is `-o` *not* supported for `nix flake prefetch`, but fetching it to thee Nix storee and trying to edit it there (or the symlink) defeats the entire purpose of an edit subcommand. As shared befeore, the goal here was to edit the given flake **directory**, not a store copy of it. This should have been as easy as picking the correct variable (or its nix2 equivalent) from NH_FLAKE, NH_OS_FLAKE etc. and opening the directory with `$EDITOR`. Conveniently nix appears to have a `nix edit` command (for whatever reason) which you can use to get rid of some of th code.
3. The blocking command implmentation and *especially* `nix flake prefetch` is an incredibly slow operation. Did you ever test this implementation?

Since the goal of the edit subcommand is to provide backwards compatibility with `nixos-rebuild edit`, I would encourage you to review `nixos-rebuild-ng` and see how they handle the edit subcommand. Do note, however, that this PR needs a serious makeover for it to be mergable.

src/nixos.rs
midischwarz12: This felt wrong because I'm trying to delete the symlink and not the actual directory which is read only. The default behavior of remove_file is to only delete the symlink when it is a link.

midischwarz12:
Unfortunate, but respectfully, I think you're a bit confused for part of this.
1. We previously talked about delaying generic platform logic to a future PR. I kind of agreed since there's a crap ton of other stuff between all three platforms that needs to be made generic other than this edit feature. Moreover, I think that PR would be best made post 4.2.0 release and maybe better made in a follow up release like a patch update since it is solely a refactor without any level of feature change. 
2. Let me break this down here:
    1. Yes, I understand that edit provides not so much value when it is a store path. However, this is the behavior of `nix edit` and `nixos-rebuild edit` for non-path flake attributes. Maybe I should parse the attribute and do the same as I am for `Installable::File` as that actually follows the path and opens it directly.
    2. Before I continue, yes, I've tested all of the os platform code. All Darwin and home implementations are untested as I don't use either. But I figured you don't mind since we previously stated that we don't test it anyways. Plus, they're *nearly* identical with only slightly discrepancies. Although, I'm really not sure why you're asking if I tested this when I marked this as tested. Did you test this and somehow failed? If so, what did you run that failed for you? Can you show the exact arguments you ran? 
    3. No, `-o` and `--outpath` are both supported by `nix flake prefetch` and can be found in the manual. I've even tested this on the command line and in the this PR. It prefetches the flake and creates a symlink to `result`. I thought it would be reasonable to create an outpath since I need the store path after prefetching. Only other way I figured possible would be to parse stdout for the store path, but this isn't very future proof, easily susceptible to bitrot, and felt a bit hacky.
    4. For `configuration.nix`, you'd have to use `--file` because this is legacy and not a flake. Although, I guess I should try to determine if it is a flake or not when using `--file` or leave if only for legacy behavior. What do you think is more reasonable?
    5. Huh, I could have sworn `nix edit` only worked for `packages` and `legacyPackages`. Maybe there was some other reason I didn't like it. I'll get back to you if I remember. 
    6. Ha, I think i undid too far at one point because I could have sworn I removed `nano` as the default editor in this change as previously discussed in this PR that it was not appropriate and should instead fail with a sensible error when `EDITOR` is unset. I'm not sure what happened, but I'll add that back in. 
3. What are you referring to by "blocking command" implementation?

Yeah, I'll look into maybe reusing the existing commands again, parse the flake attribute for paths, and add *again* the sensible error for when EDITOR is unset.

midischwarz12:
Ah, okay. I remember why I don't like nixos-rebuild edit now. This failed o my flake with:
```
error: cannot find positional information for <flake>
```

I think I was thinking that this fails on flakes which generate their nixosConfigurations output from some subsequent function. So for example, if you many hosts and have some sort of function which loops through the contents to automatically generate this (which is a very common pattern), it would fail.

However, I tried this on another flake which does not do this. And I am receiving the same error. So I'm not really sure why I'm getting this now. 

I looked at nixos-rebuild-ng and apparently they just use `nix edit` but this led me to the definition of `runCommand` for my flake. So not very optimal either.

midischwarz12:
I hate to look stupid here. But I just did some light poking around with `nixos-rebuild edit`. And it seems that it does not respect `--flake` at all. So I'd either have to go with `nix edit` somehow or completely come up with my own solution here.

midischwarz12:
@NotAShelf This should be what you want while being backwards compatible. Feel free to test and merge.

src/darwin.rs
```diff
@@ -231,3 +234,86 @@ impl DarwinReplArgs {
...
+                            .show_output(true)
+                            .run()?;
+
+                        std::fs::remove_file("result")?
+                    }
+                };
+            }
+            Installable::Store { .. } => bail!("Nix doesn't support nix store installables."),
+            Installable::Expression { .. } => bail!("Nix doesn't support nix expression."),
+        };
```
faukah, line 315: These errors could be more informative

src/interface.rs
```diff
@@ -468,6 +507,9 @@ pub enum HomeSubcommand {
 
     /// Load a home-manager configuration in a Nix REPL
     Repl(HomeReplArgs),
+
+    /// Edit home-manager configuration
```
faukah:
```suggestion
    /// Edit a home-manager configuration
```

faukah (commented):
Bunch of code duplication, but you guys already talked about that.

src/darwin.rs
midischwarz12: @faukah I'm not sure what other to say than nix cannot go and edit a store path or expression. If this is the case, nh * repl and I think a few others should change their errors too. Because I wrote this to stay consistent with the rest of the codebase.

src/interface.rs
midischwarz12: Fair

Issue #337, opened 2025-06-27
Possible race condition


timon-schelling:
I encountered an issue when running `nh os boot`. The first run resulted in the following error:

```
> Building NixOS configuration
Finished at 15:24:19 after 0s
Error:
   0: Target profile path does not exist: /tmp/nh-osGY12HT/result

Location:
   src/nixos.rs:196
```

Immediately after, I ran the same command again and everything worked as expected. I was unable to reproduce the error afterward.

This could potentially be a race condition or a timing issue, but I'm not certain.

I realize this isn’t very actionable without a reproducer. Please feel free to close if it’s not worth investigating further, but I thought it might be useful to report just in case it points to something deeper.

NotAShelf:
This has been occuring in a few places as of late. I'm not exactly sure where the race condition pops up, but it *did* come up and it's very much a bug.

acaloiaro:
I wonder if this is related, when doing `nh home switch` (`nh home build` succeeds)

```
DEBUG Configured envs: HOME=/home/adriano, NH_CURRENT_COMMAND=home, USER=adriano, PATH=<preserved>, NIX_PATH=<preserved>, LOCALE_ARCHIVE=<preserved>, NH_FLAKE=/home/adriano/git/nixos-system (nh/src/commands.rs:187)
> Activating configuration
DEBUG cmd=Exec { /tmp/nh-homeNV4KFM/result/activate } (nh/src/commands.rs:352)
Error: 
   0: Activation failed
   1: Activating configuration (exit status Exited(1))

Location:
   src/commands.rs:369
```

I can confirm that the above path `/tmp/nh-homeNV4KFM` does not exist, which explains why activation fails.

What I'm not sure how to debug is why `/tmp/nh-homeNV4KFM` doesn't exist. Forgive me, I'm not familiar with how nh works and simply wanted to contribute my own findings. I imagine this directory should be the result of `nh home build` success. The result of `ls /tmp | grep ^nh-home` is also empty.

This is failing reliably for me (I can't switch to a new home generation), so if there are any debug steps I can help with, send some commands my way and I'm happy to help. 

Versions
> nh 4.1.2
> nix (Nix) 2.28.4

[edit]

I'll make updates here as I attempt to debug. 

- I see the syscall to create a `/tmp/nh-home****` directory when running `strace -e trace=mkdir nh home build` -> `mkdir("/tmp/nh-home13RaVb", 0777)       = 0`
[...]

PR #386, opened 2025-08-16
treewide: apply clippy lints

NotAShelf:
We've previously used `Box<dyn MaybeTempPath>` to manage temporary output paths. When these boxed objects went out of scope, the embedded `TempDir` could be dropped early, causing the temporary directory to be deleted before activation scripts executed. Hence, the "race condition". The words "could be" are doing a lot of heavy lifting here, but I *think* I pinpointed the cause.

The solution to the problem, although I don't quite like it, is to replace the boxed trait object pattern with explicit tuples. Now:

- `PathBuf` contains the actual path to use
- `Option<TempDir>` keeps the temporary directory alive until function end
- When `out_link` is provided, no tempdir is created (`None`)
- When no `out_link`, a tempdir is created and held in `Some(dir)`

Hopefully this ensures that the temp dir remains vailable throughout the entire activation process. Therefore this should resolve activation failures for users experiencing the race condition, particularly on faster systems where the timing window was more likely to occur.

CC @faukah for a quick review since this partially affects Dix diffing.

Signed-off-by: NotAShelf <raf@notashelf.dev>
Change-Id: I6a6a69649e1c7cce5b78700c123a5cf3059ce4bc<!--
---


NotAShelf:
viper's fault btw :trollface:

faukah (approved):
Looks fine to me.

acaloiaro:
@NotAShelf this fix is effective, even if you don't love it. 

I recreated the conditions I posted about in #337, and with this fix it didn't have any activation issues.

NotAShelf:
Thank you for letting me know, merging after I add the changelog entry. With this 4.2.0 should be ready :D

PR #382, opened 2025-08-09
treewide: replace dialoguer with inquire

NotAShelf:
Better maintained apparently.

This replaces the dialoguer library that we've used for prompting users for confirmation with the [inquire](https://docs.rs/inquire/latest/inquire/) library. Aside from being slightly better maintained, it appears to have solved signal handling issues that dialoguer had.

Change-Id: I6a6a6964ec4608a2840c54bae939c692fd747049

Issue #203, opened 2024-12-29
Ask for sudo on the start


Acisama:
Is there some way to make it so it asks for the password at the start of the upgrade?

Because if you download many things it takes a while and sometimes I just want to do other stuff during that time, but sudo times out after 30 seconds.
Or can I use some other method to achieve this?

viperML:
`nh os switch --ask` or `sudo nh os switch -R`

viperML:
The notification should be triggered for long builds (e.g longer than 60s) and use some cross-platform crate that supports darwin

HirschBerge:
Not sure if you were actually asking for recommendations, or just jotting it down for future, but [notify](https://github.com/hoodie/notify-rust) seems to fit the criteria.

Specifically mentions that it was originally targeted for darwin. I've used it in small personal projects and it works well.

viperML:
PR's are welcome

HirschBerge:
Can you point me in the direction of the correct portion of the code base. I can take a look when I have time.

CuboidRaptor:
Would love this, especially as just sudo'ing it is unsupported

hellconnon:
I also would love this :) a notification would be neat, since i personally have 0 xp with rust i think i would break it trying to do it ^^

Yeshey:
This would be nice, I have been doing `sudo echo ; nh os switch` to get around this

NotAShelf:
I think I might be able to add a --notify flag (and NH_NOTIFY to set the flag via env) so that it sends notifications when
 
1. Sudo password is prompted
2. the switch command is done

Alternatively I can switch this into two environment variables (NH_SUDO_NOTIFY and NH_SWITCH_NOTIFY or similar), which would be both implied by the `--notify` flag. I understand that this issue is about notifications for sudo *specifically*, however, there might be more value in a fine-grained notification system in order to justify the dependency cost of something like the `notify` crate. Feedback is welcome, I'll try to submit a PR this evening for those willing to test.

PR #387, opened 2025-08-16
various: system notifications

r0chd:
Implemented system notifications with dbus notification spec, notification is sent right before
password prompt, also implemented `--ask` flag with notification actions

its currently implemented only for nixos, looking for some input before implementing more for interfaces

image

image

## Sanity Checking

- Style and consistency
- Correctness
        logic
        description.
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

## Summary by CodeRabbit

* **New Features**
  * --ask becomes tri-state (prompt, notify, both); flag without value defaults to prompt.
  * Commands can send desktop notifications to request or drive confirmations.
  * Informational and warning messages may also appear as desktop notifications alongside terminal output.

* **Chores**
  * Added a desktop notification backend and integrated it into logging and confirmation flows.

### DIFF
--- a/src/nixos.rs
+++ b/src/nixos.rs
@@ -14,6 +14,7 @@ use crate::{
   interface::{
     self,
     DiffType,
+    NotifyAskMode,
     OsBuildVmArgs,
     OsGenerationsArgs,
     OsRebuildArgs,
@@ -22,6 +23,7 @@ use crate::{
     OsSubcommand::{self},
   },
   nh_info,
+  notify::NotificationSender,
   update::update,
   util::{ensure_ssh_key_login, get_hostname, print_dix_diff},
 };
@@ -39,7 +41,7 @@ impl interface::OsArgs {
       OsSubcommand::Test(args) => args.rebuild(&Test, None, elevation),
       OsSubcommand::Switch(args) => args.rebuild(&Switch, None, elevation),
       OsSubcommand::Build(args) => {
-        if args.common.ask || args.common.dry {
+        if args.common.ask.is_some() || args.common.dry {
           warn!("`--ask` and `--dry` have no effect for `nh os build`");
         }
         args.rebuild(&Build, None, elevation)
@@ -132,7 +134,7 @@ impl OsRebuildArgs {
     self.handle_dix_diff(&target_profile);
 
     if self.common.dry || matches!(variant, Build | BuildVm) {
-      if self.common.ask {
+      if self.common.ask.is_some() {
         warn!("--ask has no effect as dry run was requested");
       }
 
@@ -317,10 +319,22 @@ impl OsRebuildArgs {
   ) -> Result<()> {
     use OsRebuildVariant::{Boot, Switch, Test};
 
-    if self.common.ask {
-      let confirmation = inquire::Confirm::new("Apply the config?")
-        .with_default(false)
-        .prompt()?;
+    if let Some(ask) = &self.common.ask {
+      let confirmation = match ask {
+        NotifyAskMode::Prompt => {
+          inquire::Confirm::new("Apply the config?")
+            .with_default(false)
+            .prompt()?
+        },
+        #[cfg(all(unix, not(target_os = "macos")))]
+        NotifyAskMode::Notify => {
+          NotificationSender::new(
+            "nh os switch",
+            "Do you want to apply the NixOS configuration?",
+          )
+          .ask()
+        },
+      };
 
       if !confirmation {
         bail!("User rejected the new config");
@@ -465,13 +479,25 @@ impl OsRollbackArgs {
       return Ok(());
     }
 
-    if self.ask {
-      let confirmation = inquire::Confirm::new(&format!(
-        "Roll back to generation {}?",
-        target_generation.number
-      ))
-      .with_default(false)
-      .prompt()?;
+    if let Some(ask) = &self.ask {
+      let confirmation = match ask {
+        NotifyAskMode::Prompt => {
+          inquire::Confirm::new(&format!(
+            "Roll back to generation {}?",
+            target_generation.number
+          ))
+          .with_default(false)
+          .prompt()?
+        },
+        #[cfg(all(unix, not(target_os = "macos")))]
+        NotifyAskMode::Notify => {
+          NotificationSender::new(
+            "nh os rollback",
+            &format!("Roll back to generation {}?", target_generation.number),
+          )
+          .ask()
+        },
+      };
 
       if !confirmation {
         bail!("User rejected the rollback");
@@ -659,7 +685,7 @@ fn print_vm_instructions(out_path: &Path) {
   }
 }
 
-/// Runs the built NixOS VM by executing the VM runner script.
+/// Runs the built `NixOS` VM by executing the VM runner script.
 ///
 /// Locates the VM runner script in the build output directory and executes it,
 /// streaming its output to the user. Returns an error if the script cannot be
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,8 +38,8 @@ hostname = "0.4.1"
 humantime = "2.3.0"
 inquire = { default-features = false, version = "0.9.1", features = [ "crossterm" ] }
 nix = { default-features = false, features = [ "fs", "user" ], version = "0.30.1" }
-regex = "1.11.3"
 notify-rust = "4.11.7"
+regex = "1.11.3"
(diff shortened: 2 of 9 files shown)


NotAShelf:
I'd rather if this was done with the notify crate and not zbus. Zbus has a history of holding critical fixes for months, and I'd rather avoid Linux-only features since it turns out as future maintenance work one way or another.

Also on a more personal note, this seems to be a bit too much. NixOS icons dependency? zbus? Having to switch between notifications or terminal prompt? Rather drastic imo. We should be able to send a notification using the aforemetioned crate with much less code, including with actions using libnotify.

r0chd:
notify crate makes sense to me, but it'd still depend on zbus, unless you prefer dbus-rs backend (but that would require dependency on libdbus and pkg-config).

The terminal prompt was just a fallback in case daemon wasnt available or notification was dismissed/expired, I'm not sure how to handle it in a better way.

Also do you think the notification should still use nixos-snowflake icon but just get rid of the dependency or ditch icons all together?

NotAShelf:
We could inline a SVG for the icon, probably preferable.

src/nixos.rs
```diff
@@ -233,17 +248,57 @@ impl OsRebuildArgs {
...
+                                .prompt()
+                                .unwrap_or_default(),
+                        }
+                    });
+                    confirmation
+                }
+                Err(_) => inquire::Confirm::new("Apply the config?")
+                    .with_default(false)
+                    .prompt()?,
+            };
```
NotAShelf: I have given this some thought, and I think I prefer the old approach for the prompt. Namely, I think the ask behaviour should remain as-is unless thee user explicitly opts into notification prompts. Otherwise this will be a very invasive change that I, for one, would not appreciate in my setup. Also note that using `info!("Apply the config?");` and ` inquire::Confirm::new("Apply the config?")` will actually prompt the user twice. We had previously used `info` for the question, and the input prompt would not display a message. This has changed with #382.

Imo the correct approach here would be to simply add a `--notify` flag that coexists with the prompt. Without having checked myself, I think it would be a nice addition to make it so that if `--notify` flag is available (or maybe something along the lines of `--ask=notif`, `--ask=prompt` and `--ask=both` in addition to `--ask` that would imply `--ask=prompt` or similar for backwards compat) the notification pops up with the confirmation prompt *alongside* the terminal prompt. Though if you think this is too much, then I'd prefer a simple notification for when

1. A confirmation prompt is available
2. A task is done (e.g., a rebuild)

 This can also be used to notify the user of an awaiting sudo prompt if the logic can be made a bit more generic, but that would have to wait until #92 is merged. See [here](https://github.com/nix-community/nh/issues/203#issuecomment-3166570126) for my previous comment on the matter.

r0chd: Flags look good to me, I could wait until #92 as currently notify-rust has a weird behavior with platform specific options where instead of being a noop they're not included in the build https://github.com/hoodie/notify-rust/issues/223 so we're gonna end up with a lot of ugly cfg directives and weird branching, I'm gonna try and merge a PR before doing anything else with this change

NotAShelf: #92 has now been merged.

I'm aiming for a new release *at worst* on Monday, but if you think that won't be too much of a Rush and you can finish the refactor by then I can try to do it Monday afternoon or on Tuesday morning. Otherwise this'll have to be for the 4.3 release.

I've never used notify-rust before (I was thinking of the other notify crate, for some reason) but waiting for the bug you've linked to be fixed might take too long. If possible, consider alternative libraries.

r0chd: I probably won't be able to find time to finish it before the release

NotAShelf:
Actual yapmachine holy hell

r0chd:
https://github.com/nix-community/nh/pull/387/files#diff-fb0614459ea1fee2430f5df8e36530289fa54b04e088be756d2db2054c8d7081R91-R136

@NotAShelf Is this too much voodoo for our purposes, for our mission statement?

src/generations.rs
```diff
@@ -382,11 +382,11 @@ pub fn print_info(
...
           Field::Size => generation.closure_size.clone(),
         };
-        format!("{:width$}", cell_content)
+        format!("{cell_content:width$}")
       })
       .collect::<Vec<String>>()
       .join(" ");
-    println!("{}", row);
+    println!("{row}");
   }
```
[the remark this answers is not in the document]
NotAShelf: Stupid robot, this is false

leiserfg:
Couldn't this be done with a simple `OSC-99` so nh does not have to contain a dbus client just for a notification?

NotAShelf:
Thanks for taking the time to work on this, @r0chd. I appreciate the level of effort that clearly went into the PR. The scope has grown quite a bit compared to the original intention, and that makes it harder to evaluate confidently. We've had situations in the past where increased complexity didn't pay off in terms of maintainability, and I want to avoid repeating that pattern.

In addition, several other PRs have recently been merged that touch large parts of the codebase, so this branch will need a rebase before it can move forward.

Before reviewing in depth, I'd like to ask for another pass over the implementation with an eye toward simplification. As @leiserfg pointed out, an OSC99-based approach might give us the same functionality while keeping both the crate size and the conceptual load under control. If there are parts of the current design that can be streamlined or reduced, that would make the review process significantly easier and improve the long-term maintainability of the feature.

I am, of course, open to discussing alternatives if needed, but a simplified iteration would be a solid next step.

JonnieCache:
FWIW claude code uses in-terminal control code notifications (presumably OSC99,) that being a popular product it's likely to be well-supported by terminal clients. It uses them for precisely this purpose, when it's waiting for input after a long task, and clicking on them switches my desktop to reveal the window as well.

I believe there are even control codes to detect if the terminal window has focus/is visible?

r0chd:
I wasn't aware of OSC-99 before, at first glance it does look like a way better way, but it has a major problem, I wasn't able to find a terminal besides kitty that supports the escape sequence (ghostty seems to plan on implementing it tho)

generally I agree that the escape sequence should be used instead of depending on dbus, just let me know if you'd prefer to stay with the current impl, use escape sequence or just ditch the feature all together

P.S. if you have something to add, please tag me, I assume that all notifications from this pr are spam from the bot and just ignore them

NotAShelf:
@coderabbitai shut up

NotAShelf:
That suppresses the roomba.

> I wasn't aware of OSC-99 before, at first glance it does look like a way better way, but it has a major problem, I wasn't able to find a terminal besides kitty that supports the escape sequence (ghostty seems to plan on implementing it tho)

From a quick glance, Kitty and Foot seem to be the ones that support it. I don't know if Alacritty does, but I generally dislike Alacritty and how it is maintained so I'm happy to tell people to use a better terminal instead of supporting it. I would also expect most DE-specific terminals such as gnome-terminal and konsole to support them but you're right, it is not too wide-spread.

> generally I agree that the escape sequence should be used instead of depending on dbus, just let me know if you'd prefer to stay with the current impl, use escape sequence or just ditch the feature all together

I'm guessing the current impl will be fine, but I'd like for it to be simplified. While I haven't looked at a better solution myself, I would *hope* that adding notifications (dbus or not) is a trivial task.

RE: @r0chd

leiserfg:
There is also OSC 9, which has wider adoption but is less feature-full (one can only set the message of the notification). Alacrity does not support it either, but if some one uses that term, they does not care about features.

r0chd:
> From a quick glance, Kitty and Foot seem to be the ones that support it. I don't know if Alacritty does, but I generally dislike Alacritty and how it is maintained so I'm happy to tell people to use a better terminal instead of supporting it. I would also expect most DE-specific terminals such as gnome-terminal and konsole to support them but you're right, it is not too wide-spread.

tried running it in foot, kitty, ghostty, gnome-terminal, konsole and alacritty and only kitty worked for me, ghostty _planning_ to implement it imo is good enough for it not to feel like a kitty specific feature. Could someone else test OSC-99 support in those terminals? maybe I'm just stupid

eveeifyeve:
> That suppresses the roomba.
XD

> tried running it in foot, kitty, ghostty, gnome-terminal, konsole and alacritty and only kitty worked for me, ghostty planning to implement it imo is good enough for it not to feel like a kitty specific feature. Could someone else test OSC-99 support in those terminals? maybe I'm just stupid

It might be worth writing a test to test in terminals as a nix test or cargo test.

PR #311, opened 2025-06-03
Add Nix Build Passthrough Arguments to Rebuild Commands

cdo256:
This patch adds support for passing Nix build flags (like --max-jobs, --keep-going, --show-trace, etc.) directly to rebuild commands. A new NixBuildPassthroughArgs struct collects these options and passes them through the CLI to the build logic for Darwin, NixOS, and Home Manager workflows.

The code is based on the [consolidate-logic](https://github.com/nix-community/nh/commits/consolidate-logic). I have cherry picked Andreas Fuchs commit so it runs on the version of Lix I use. Feel free to disregard that and merge it later if easier.

I'm open to feedback about which arguments to actually pass through. Other than `-t` for `--show-trace`, these are all standard. 

Currently I have:

```rust
#[derive(Debug, Args)]
pub struct NixBuildPassthroughArgs {
    /// Number of concurrent jobs Nix should run
    #[arg(long, short = 'j')]
    pub max_jobs: Option<usize>,

    /// Number of cores Nix should utilize
    #[arg(long)]
    pub cores: Option<usize>,

    /// Logging format used by Nix
    #[arg(long)]
    pub log_format: Option<String>,

    /// Continue building despite encountering errors
    #[arg(long, short = 'k')]
    pub keep_going: bool,

    /// Keep build outputs from failed builds
    #[arg(long, short = 'K')]
    pub keep_failed: bool,

    /// Attempt to build locally if substituters fail
    #[arg(long)]
    pub fallback: bool,

    /// Repair corrupted store paths
    #[arg(long)]
    pub repair: bool,

    /// Explicitly define remote builders
    #[arg(long)]
    pub builders: Option<String>,

    /// Paths to include
    #[arg(long, short = 'I')]
    pub include: Vec<String>,

    /// Print build logs directly to stdout
    #[arg(long, short = 'L')]
    pub print_build_logs: bool,

    /// Display tracebacks on errors
    #[arg(long, short = 't')]
    pub show_trace: bool,

    /// Accept configuration from flakes
    #[arg(long)]
    pub accept_flake_config: bool,

    /// Refresh flakes to the latest revision
    #[arg(long)]
    pub refresh: bool,

    /// Allow impure builds
    #[arg(long)]
    pub impure: bool,

    /// Build without internet access
    #[arg(long)]
    pub offline: bool,

    /// Prohibit network usage
    #[arg(long)]
    pub no_net: bool,

    /// Recreate the flake.lock file entirely
    #[arg(long)]
[...]

PR #397, opened 2025-08-25
feat: add --refresh to darwin/home/os

lovesegfault:
It's really common when working with remote flakes to always want the
latest upstream ref, and not the local cached one, which is what
`--refresh` is for.

It'd be nice if it was passable directly in `nh` instead of as an extra
arg.

## Sanity Checking

- Style and consistency
  - [x] I have added appropriate docunentation to new code
- Correctness
  - [ ] My code includes comments in particularly complex areas to explain the
        logic
        description.
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ functionality, under the "Removed" section.
   - It's roughly %4 faster according to testing, but IO is still a limiting
     factor and results may differ.
 - Added more context to some minor debug messages across platform commands.
+- The `--refresh` flag is now treated as a passthrough argument to Nix.
 
 ### Fixed
 
--- a/src/darwin.rs
+++ b/src/darwin.rs
@@ -102,13 +102,19 @@ impl DarwinRebuildArgs {
 
         let toplevel = toplevel_for(hostname, processed_installable, "toplevel");
 
-        commands::Build::new(toplevel)
+        let mut build = commands::Build::new(toplevel)
             .extra_arg("--out-link")
             .extra_arg(&out_path)
             .extra_args(&self.extra_args)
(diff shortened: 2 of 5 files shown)


NotAShelf:
Is there a reason this cannot be a passthru arg? Since it's a native flag in Nix, and since there is no additional logic on NH's end I think we can simply relay it to Nix as is.

lovesegfault:
@NotAShelf not at all, that's a great suggestion, pushed those changes :)

lovesegfault:
Ah, nevermind, I misunderstood how those work. @NotAShelf, I wanted to avoid having to do `nh os switch whateverFlake -- --refresh` every time, and instead having the ability to write `nh os switch --refresh whateverFlake` because I essentially use `--refresh` on _every_ deploy.

If you feel it's not worth it, I understand :)

NotAShelf:
> I wanted to avoid having to do nh os switch whateverFlake -- --refresh every time, and instead having the ability to write nh os switch --refresh whateverFlake because I essentially use --refresh on every deploy.

Passthrough args allow you to pass Nix commands directly to nh commands without the `-- <flag>` syntax. Normally you'd run `nh os switch -- --refresh` while in passthrough it becomes `nh os switch --refresh`. You can see this in action for each passthrough flag, e.g., `--builders`.  

See the relevant PR: #311

lovesegfault:
Ah, I see why I was confused, this hasn't been released yet and it's not in the `CHANGELOG`. If I run `nh` off `master` it just works. Thanks!

PR #411, opened 2025-09-07
commands: ask for SSH password locally for remote deployments

NotAShelf:
Improves the remote deployment experience by prompting for a `sudo` password when deploying to remote hosts, rather than requiring password-less `sudo`. 

During testing I've noticed that the sudo password is prompted **3 times** (6 if we enable Inquire's confirmation) so we also introduce a password cache to avoid repeated prompts. Changelog and tests have been updated accordingly.

## Summary by CodeRabbit

* **Changed**
  * Remote deployments now prompt for a sudo password when elevation is required, reducing reliance on passwordless sudo on target hosts.
  * Per-host password caching for the session minimizes repeated prompts during multiple operations.
  * Remote elevation flow improved for smoother, more reliable deployments over SSH.

* **Documentation**
  * Changelog updated to describe the new remote sudo prompt behavior and clarify interaction with existing privilege escalation options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,8 +50,8 @@ functionality, under the "Removed" section.
   `run0` and a fallback `pkexec` strategies will be attempted if the system does
   not use `sudo`.
 - Nh will correctly prompt you for your `sudo` password while deploying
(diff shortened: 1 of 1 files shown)


NotAShelf:
What the hell

Issue #375, opened 2025-08-04
fine-grain `nh os info; add variants for `nh home` and `nh darwin`


NotAShelf:
Currently `nh os info`

1. Hard-codes the profile path
2. Displays all information at once (fixed)
3. Only shows stats for NixOS installations

Would be nice to take a `--profile` argument like `nix-env --list-generations --profile` does to allow specifying, mostly for backwards compatibility. It would also be nice to allow specifying which fields to display. Namely, we want to allow filtering for **name**, **closure size** and **generations list**. Display all when no specifier passed.

PR #413, opened 2025-09-08
nh os info allow specifying which field to display

cloud-yu:
Hi, 
for https://github.com/nix-community/nh/issues/375 , I make some changes for `nh os info`

1. add a closure size column for every generations 
image

2. add a args --fmt to specify which field to display
image

## Sanity Checking

- Style and consistency
  - [ ] I have added appropriate docunentation to new code
- Correctness
        logic
  - [ ] I have documented the motive for those changes in the PR body or commit
        description.
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---


src/generations.rs
```diff
@@ -45,9 +49,35 @@ pub fn from_dir(generation_dir: &Path) -> Option<u64> {
...
 }
 
+#[must_use]
+pub fn get_closure_size(generation_dir: &Path) -> Option<String> {
+    match (Exec::cmd("nix")
+        .arg("path-info")
+        .arg(generation_dir)
+        .arg("-Sh")
+        .arg("--json")
+        | Exec::cmd("jq").arg(".[]"))
```
NotAShelf: Relying on a shell tool for JSON parsing is unacceptable. We already have serde_json, use it please. I'm also not sure why this uses `Exec` now.

src/interface.rs
```diff
@@ -330,6 +330,19 @@ pub struct OsGenerationsArgs {
...
+    /// field:
+    ///     id: generation id
+    ///     date: build date
+    ///     nver: nixos version
+    ///     kernel: kernel version
+    ///     confRev: configuration revision
+    ///     spec: specialisation
+    ///     size: closure size
+    #[arg(long, value_delimiter = ',', verbatim_doc_comment)]
+    pub fmt: Option<Vec<String>>,
```
NotAShelf: This should be `fields` and not `fmt` imo.

src/generations.rs
cloud-yu: use `Exec` just for the pipeline operation.

`nix path-info -S --json` returns a map not an array; parse the first element with `jq[0]` get `Value::Null`
I'll rollback to `Command`, and add get the realpath with `read_link` for the map index name.

image

src/interface.rs
cloud-yu: I prefer `fmt` because it also support ordered the table header.
image

src/generations.rs
NotAShelf: Please don't use jq for this.

src/interface.rs
NotAShelf: I still think fields makes more sense here. It does not imply an ordering but the name of the flag itself makes more sense. With something like fmt I'd assume I can customize more than just the fields to be displayed.

cloud-yu: Ok, I'll fix it

src/interface.rs
```diff
@@ -345,6 +346,15 @@ pub struct OsGenerationsArgs {
   /// Path to Nix' profiles directory
   #[arg(long, short = 'P', default_value = "/nix/var/nix/profiles/system")]
   pub profile: Option<String>,
+
+  /// Specify which field(s) to show (comma-delimited).Defaults to show all
+  /// fields.
```
NotAShelf:
```suggestion
  /// Comma-delimited list of field(s) to display
```

If I recall correctly, Clap already provides a hint in the help text from default_value so we'll be describing the default value twice here.

NotAShelf:
Thank you :)

Issue #620, opened 2026-04-14
nh does not show what is being copied with --target-host


rice-cracker-dev:
When switching configurations to the remote server, nh does not show what is being copied.

<img width="1009" height="140" alt="Image" src="https://github.com/user-attachments/assets/40f3f9c2-4a41-42f1-9404-05287f4373fd" />

faukah:
Should be fixed on master.

PR #626, opened 2026-04-21
Fix output visibility for nix copy commands after subprocess update

### LINKED ISSUE #624: nh clean regression in 4.3.1: paths deleted / freed summary no longer prints
### Summary

In nh 4.3.1, nh clean no longer prints the nix store gc summary line (N store paths deleted, X MiB freed). The line printed correctly in nh 4.3.0 and was reliably visible for months prior. The progress header (Performing garbage collection on the nix store) and generation cleanup still print — only the final size/freed line from the underlying nix store gc is missing.

### Reproduction

```
nh clean all --keep-since 1h
```

On a host with generations to collect, output ends at:

```
> Removing /nix/var/nix/profiles/system-XXX-link
> Performing garbage collection on the nix store
```

No N paths deleted, Y MiB freed line follows.

### Environment

- nh 4.3.1
- nix 2.34.6
- NixOS (unstable), x86_64-linux

### Cause

The subprocess dependency jumped 0.2.14 → 1.0.0 between 4.3.0 and 4.3.1 (cargo bump in 0e42e3131852). The `Command::run()` path in crates/nh-core/src/command.rs sets `stderr = Redirection::Merge` when `show_output(true)`, leaves stdout untouched (so it stays at `Redirection::None`, meaning "inherit parent stdio"), and then calls `.capture()`.

`.capture()` semantics changed across the dep bump.

**subprocess 0.2.14** (`setup_communicate`, called by `capture`):

```rust
if let (&Redirection::None, &Redirection::None) =
    (&self.config.stdout, &self.config.stderr)
{
    self = self.stdout(Redirection::Pipe);
}
```

The `None → Pipe` override only fires when *both* streams are still `None`. With `stderr = Merge`, the guard fails, stdout stays `None` (inherit), and the merged stream passes through to the terminal.

**subprocess 1.0.0** (`capture`):

```rust
if matches!(*self.stdout_redirect, Redirection::None) {
    self = self.stdout(Redirection::Pipe);
}
if matches!(*self.stderr_redirect, Redirection::None) {
    self = self.stderr(Redirection::Pipe);
}
```

Each stream is now upgraded independently. With `stderr = Merge` and stdout = `None`, stdout is forced to `Pipe` and the merged stream lands in a memory buffer that `Command::run()` never reads on success (only `capture.stderr_str()` on failure), so the nix store gc freed-size line (which nix prints to stdout) is silently dropped.

Net effect: nh was relying on the 0.2.x "both-None" heuristic for terminal pass-through. The 1.0 cleanup tightened the rule to per-stream and removed the heuristic; MIGRATION.md claims `capture()` "works the same way" but this case is a silent behavior change.

### Suggested fix

Either:

1. Use `Exec::join()` instead of `Exec::capture()` when `show_output` is set. `join()` doesn't alter redirection; stdout stays `None` = inherit, and the terminal sees output again (closest to pre-1.0 behavior).
2. Print `capture.stdout_str()` after successful execution when `show_output` is set.

Option 1 is the minimal change.

### Related

- PR #343 (v4.2.0) — added `show_output(true)` to the gc call. Worked in 4.3.0; silently broken by the subprocess 1.0 upgrade.

NotAShelf:
Add `exec_with_streaming()` helper to nh_core::command that uses `.start()` with piped stdout/stderr and spawns threads to stream output to the terminal in real-time. This captures output for error reporting on failure, and HOPEFULLY fixes the stdout/stderr bs that was happening after the Subprocess 1.0 upgrade.

Fixes: #624, #620

## Summary by CodeRabbit

* **New Features**
  * Real-time streaming of command output: stdout/stderr shown live during execution and still returned after completion.

* **Bug Fixes**
  * Remote copy failures now report process exit status consistently when streaming output is enabled.

* **Chores**
  * Workspace dependency for subprocess updated to v1.0.3 to align dependency resolution across the project.


NotAShelf:
I don't like the helper approach btw, but it's the cleanest way we can do this in a futureproof manner imho. Unless we decide to pull `nh-subprocess` out of somewhere the sun don't shine.

faukah (approved):

PR #647, opened 2026-05-16
nh-clean: cleanup; improve UX

NotAShelf:
1. Replaces loose `RESULT_REGEX` with `is_nix_store_direct_child()`, which checks for `/nix/store` direct children structurally. 
2. Fixes orphaned gcroot detection; missing dst now tags src before the filter runs, and broken symlinks whose store-path target was GC'd are tagged unconditionally. 
3. Also replaces the flat `read_dir` of `/auto` with a `WalkDir` over all of `/nix/var/nix/gcroots`, gated by `--cross-filesystems/-x` for filesystem boundary control.

Largely inspired by <https://github.com/blitz/nix-link-cleanup>. One solution I haven't fixed is non-standard Nix store paths. AFAIK Nix can have the Nix store at a non-standard path outside of `/nix` and while most systems will not do this, it's an issue I haven't tackled. There is no good way to get the correct path without parsing `nix.conf`, which I don't want to do quite honestly.

Change-Id: I8265c8f4dfc6aea938a0739f47e1fe476a6a6964

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,30 @@ functionality, under the "Removed" section.
 
 ## Unreleased
 
+### Added
+
+- `nh clean --keep-one` preserves all active direnv gcroots regardless of
+  `--keep-since`, preventing projects from being collected when they haven't
+  been activated recently. Orphaned and broken gcroots are still removed.
+- `nh clean --cross-filesystems` / `-x` allows the gcroot scan to cross
+  filesystem boundaries. By default the walk stays on the same filesystem as
+  `/nix/var/nix/gcroots`.
+
 ### Changed
 
 - `nh search` now uses search backend version 48 (previously 46) to track the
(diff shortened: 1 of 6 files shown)


qweered:
>There is no good way to get the correct path without parsing nix.conf, which I don't want to do quite honestly.

You can use `nix config show`

faukah:
> > There is no good way to get the correct path without parsing nix.conf, which I don't want to do quite honestly.
> 
> You can use `nix config show`

`nix config show --json` might actually work for this. Good idea!

faukah (approved):

NotAShelf:
> > There is no good way to get the correct path without parsing nix.conf, which I don't want to do quite honestly.
> 
> You can use `nix config show`

We use this for the experimental features check, but I'm honestly not a big fan of it. There is a bit of nuance to GC that I don't want to just discard by bumrushing.

PR #650, opened 2026-05-19
nh-remote: actually fix remote copy progress with custom spinner stre…

NotAShelf:
# FIXES THE GODDAMN SPINNER

I'll need therapy after this, but it finally works.


faukah (approved):

faukah:
lgtm. Changelog entry missing, all else is fine

NotAShelf:
Changelog entry not necessary because we already noted that we fixed it. This just fixes the fix.

PR #656, opened 2026-05-25
nh-nixos: add delete subcommand to remove and gc specific generations

d4rk:
This pull request introduces a `delete` subcommand to `nh os` allowing users to delete and garbage collect specific system generations.

### Examples Usages
- `nh os delete 80`
- `nh os delete 91 92`

The generation IDs can be obtained from `nh os info`. 

By default, it will also run `nix store gc` and `switch-to-configuration boot` to remove the boot menu of the pruned generations. These can be controlled with `no-gc` and `no-boot`.

I'm a new user of  `nh` and in my workflow I sometimes prefer to prune specific generations rather than everything older than a timestamp, so I decided to try adding support for it to `nh`.

## Sanity Checking

- [x] I have read and understood the [contribution guidelines]
- Style and consistency
  - [ ] I ran **`nix fmt`** to format my Nix code
- Correctness
        logic
        description.
- Tested on platform(s):
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

## AI Disclosure

Per the [AI Policy](https://github.com/nix-community/nh/blob/master/CONTRIBUTING.md#ai-policy), I'm disclosing that this feature was implemented with the assistance of Antigravity CLI. With that said, I've:
- reviewed the code thoroughly
- tried to ensure that it's following coding patterns and paradigms consistent with the project
- tested locally on my own system

---

### DIFF
--- a/crates/nh-nixos/src/nixos.rs
+++ b/crates/nh-nixos/src/nixos.rs
@@ -26,6 +26,7 @@ use crate::{
     self,
     OsBuildImageArgs,
     OsBuildVmArgs,
+    OsDeleteArgs,
     OsGenerationsArgs,
     OsRebuildActivateArgs,
     OsRebuildArgs,
@@ -93,6 +94,7 @@ impl args::OsArgs {
       OsSubcommand::Repl(args) => args.run(),
       OsSubcommand::Info(args) => args.info(),
       OsSubcommand::Rollback(args) => args.rollback(elevation),
+      OsSubcommand::Delete(args) => args.delete(elevation),
       OsSubcommand::BuildImage(args) => args.build_image(&elevation),
     }
   }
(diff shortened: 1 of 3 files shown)


crates/nh-nixos/src/nixos.rs
```diff
@@ -1434,3 +1436,135 @@ impl OsGenerationsArgs {
...
+        inquire::Confirm::new(&format!("Delete generation(s) {targets:?}?"))
+          .with_default(false)
+          .prompt()?;
+
+      if !confirmation {
+        bail!("User rejected the deletion");
+      }
+    }
+
+    info!("Deleting generation(s)...");
```
faukah, line 1527: It'd be nice if it told you here which generations it's about to remove.

crates/nh-nixos/src/nixos.rs
```diff
@@ -1434,3 +1436,135 @@ impl OsGenerationsArgs {
...
+    for (generation_num, path) in &paths_to_delete {
+      Command::new("rm")
```
faukah, line 1529: Let's not shell out to `rm` when we have a [great standard library](https://doc.rust-lang.org/std/fs/index.html#functions).

crates/nh-nixos/src/nixos.rs
```diff
@@ -1434,3 +1436,135 @@ impl OsGenerationsArgs {
...
+
+    if !self.no_gc {
+      Command::new("nix")
+        .args(["store", "gc"])
+        .dry(self.dry)
+        .message("Performing garbage collection on the nix store")
+        .show_output(true)
+        .with_required_env()
+        .run()?;
+    }
```
faukah, line 1565: instead of calling `nix store gc`, why not just use `nh-clean`?

faukah (commented):
Hey there! First of all, thank you for your pull request and *thank you* for the AI disclosure. <3 I do agree that this is something that nh is currently missing and that this is quite useful.

Two preliminary notes:
- I'm not *too* happy about this being yet another thing in `nh os`. In the long term, I feel like we should add some sort of "tidy" sub-command, which bundles several things like this exact feature. This is something for me and @NotAShelf to consider in the future though; Just adding this to `nh os` completely fine for this PR.
- It may make sense to add a check if _all_ of the generations actually exist before we start deleting them. I prefer "all-or-nothing" behavior more than it just failing at some point during running if the user specifies an incorrect generation number. 

I left a few review comments on things I spotted a first glance. I'll do a proper review later; I'm still not caffeinated for enough for that yet.

d4rk:
Thanks for the quick initial review! I had a few follow-up questions. I'll make changes based on responses. Thanks!

crates/nh-nixos/src/nixos.rs
d4rk: This is definitely my ignorance, but how would that work? My understanding is that `nh clean` will perform deletions based on a retention policy (`--keep`, etc) and _then_ run gc.  So a few clarifying questions:

1. Is there a way to invoke it to just have it do gc?
2. If there is, I assume we don't want to shell it out since that would be `nh` spawning itself?

It seems like potentially refactoring [this](https://github.com/nix-community/nh/blob/master/crates/nh-clean/src/clean.rs#L427-L450) snippet into `nh-core` and invoking it is one approach, but wondering if that's overkill :)

crates/nh-nixos/src/nixos.rs
d4rk: Is there a way to use `remove_file` with privilege escalation? This removal, at least on my system, requires sudo. 

I tested it by swapping the `Command` with a simple `fs::remove_file(path).context("Failed to remove generation")?;`, and if failed. Apologies if I missed the obvious approach.

```
> Deleting generation(s)...
Error: 
   0: Failed to remove generation
   1: Permission denied (os error 13)

Location:
   crates/nh-nixos/src/nixos.rs:1529
```

Thinking about it some more, another option is to invoke `nix-env --delete-generations`, also wrapped in an elevated `Command`. Something like:

`nix-env -p /nix/var/nix/profiles/system --delete-generations 90 91`

But your FAQ notes that you're trying to move away from shelling out to nix CLI when possible, so I haven't made the change yet. Will await feedback.

d4rk:
@faukah Friendly ping. I had 2 follow-up comments and was hoping to get your thoughts. Thanks!

crates/nh-nixos/src/nixos.rs
faukah: Hmmm. After looking around a bit, it actually seems that your approach is the better one, my bad on that. `nix-env --delete-generations` seems to be exactly what we want here, let's use that. I imagine it's gonna cause less problems in the long run than just `rm`-ing stuff.

crates/nh-nixos/src/nixos.rs
faukah: > It seems like potentially refactoring [this](https://github.com/nix-community/nh/blob/master/crates/nh-clean/src/clean.rs?rgh-link-date=2026-05-25T17%3A01%3A23Z#L427-L450) snippet into nh-core and invoking it is one approach, but wondering if that's overkill :)

Something like that *would* be nice, but to not have this PR stalled even more, I'll so so myself afterwards.

PR #683, opened 2026-06-10
nh-remote: move from nix-copy-closure to nix copy

faukah:
Move from `nix-copy-closure` to `nix copy`. We now also explicitly use `ssh://` if the user supplies it, otherwise use `ssh-ng://` whereever possible. Furthermore, we clean up the semantics of copying between hosts by adding a `CopyDirection` struct:
```
enum CopyDirection<'a> {
  FromRemote(&'a RemoteHost),
  ToRemote {
    host:            &'a RemoteHost,
    use_substitutes: bool,
  },
  BetweenRemotes {
    from_host:       &'a RemoteHost,
    to_host:         &'a RemoteHost,
    use_substitutes: bool,
  },
}
```

This should hopefully clean `nh-remote` up a bit.
Since this touches remote store handling, I'd advise testing this properly; especially copying between machines. I did do some testing myself but do not have two machines to test copying from one machine to another. 

### AI disclosure
I *did* throw Codex against at diff after implementing this, and it found a few valid problems which I fixed myself. I take full responsibility for any code in this PR.  

## Sanity Checking

- [X] I have read and understood the [contribution guidelines]
- [X] I have updated the [changelog] as per my changes
- [X] I have tested, and self-reviewed my code
- Style and consistency
  - [X] I ran **`nix fmt`** to format my Nix code
  - [X] I ran **`cargo fmt`** to format my Rust code
  - [X] I have added appropriate documentation to new code
  - [X] My changes are consistent with the rest of the codebase
- Correctness
  - [X] I ran **`cargo clippy`** and fixed any new linter warnings.
  - [X] My code includes comments in particularly complex areas to explain the
        logic
  - [X] I have documented the motive for those changes in the PR body or commit
        description.
- Tested on platform(s):
  - [X] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/crates/nh-remote/src/copy.rs
+++ b/crates/nh-remote/src/copy.rs
@@ -1,0 +1,442 @@
+use std::{
+  ffi::OsString,
+  io::{BufRead, Read},
+  path::Path,
+  time::Duration,
+};
+
+use color_eyre::{
+  Result,
+  eyre::{Context, eyre},
+};
+use indicatif::{ProgressBar, ProgressStyle};
+use nh_core::command::exec_with_streaming;
+use subprocess::{Exec, Redirection};
+use tracing::{debug, error, info};
+
+use super::{RemoteHost, get_flake_flags, get_nix_sshopts_env};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum CopyDirection<'a> {
+  FromRemote(&'a RemoteHost),
+  ToRemote {
+    host:            &'a RemoteHost,
+    use_substitutes: bool,
+  },
+  BetweenRemotes {
+    from_host:       &'a RemoteHost,
+    to_host:         &'a RemoteHost,
+    use_substitutes: bool,
+  },
+}
+
+impl CopyDirection<'_> {
+  fn args(self) -> Vec<String> {
+    match self {
+      Self::FromRemote(host) => {
+        vec!["copy".to_string(), "--from".to_string(), store_uri(host)]
+      },
+      Self::ToRemote {
+        host,
+        use_substitutes,
+      } => {
+        let mut args =
+          vec!["copy".to_string(), "--to".to_string(), store_uri(host)];
+        push_substitute_on_destination(&mut args, use_substitutes);
+        args
+      },
+      Self::BetweenRemotes {
+        from_host,
+        to_host,
+        use_substitutes,
+      } => {
+        let mut args = vec![
+          "copy".to_string(),
+          "--from".to_string(),
+          store_uri(from_host),
+          "--to".to_string(),
+          store_uri(to_host),
+        ];
+        push_substitute_on_destination(&mut args, use_substitutes);
+        args
+      },
+    }
+  }
+}
+
+fn push_substitute_on_destination(
+  args: &mut Vec<String>,
+  use_substitutes: bool,
+) {
+  if use_substitutes {
+    args.push("--substitute-on-destination".to_string());
+  }
+}
+
+fn store_uri(host: &RemoteHost) -> String {
+  host.nix_store_uri()
+}
+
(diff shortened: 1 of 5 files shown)


crates/nh-remote/src/copy.rs
```diff
@@ -0,0 +1,442 @@
...
+fn store_uri(host: &RemoteHost) -> String {
+  host.nix_store_uri()
+}
+
+fn build_nix_copy_command<P: Into<OsString>>(
+  direction: CopyDirection<'_>,
+  path: P,
+) -> Exec {
+  let flake_flags = get_flake_flags();
+  let mut cmd = Exec::cmd("nix").args(&flake_flags);
```
NotAShelf, line 85: `get_flake_flags()` returns flags like `--accept-flake-config`, `--extra-experimental-features flakes`, etc. Most flake flags are noise here. This originated in `copy_to_remote` and `copy_closure_between_remotes` but the centralization attempt in this PR might be a good time to address the issue if we can. Not a blocker, just wanted to point out.

crates/nh-remote/src/remote.rs
```diff
@@ -2596,64 +2459,4 @@ mod tests {
...
-    fn read(&mut self, _buf: &mut [u8]) -> std::io::Result<usize> {
-      Err(std::io::Error::new(
-        std::io::ErrorKind::BrokenPipe,
-        "simulated pipe failure",
-      ))
-    }
-  }
-
-  #[test]
-  fn test_exec_with_spinner_streaming_mixed_output_no_deadlock() {
```
NotAShelf, line 2614: This was not carried into `copy.rs`. The function under test  is identical to the old code, it was moved verbatim. The dual-pipe pattern (i.e., `stdout` + `stderr` read in separate `BufReader` threads) is inherently deadlock-prone so I think a regression here (e.g., someone adding blocking behavior before threads are joined) would now go undetected. Up to you whether to keep this, but I probably would given how fragile spinner stuff used to be.

faukah: My main reasoning for dropping this was not wanting to rely on `bash`, but I guess that makes sense. I'll put it back in.

crates/nh-remote/src/copy.rs
faukah: I'll address this in a future PR; `nh-remote` will be seeing some more work in the near future anyways. I'll keep it in mind for then.

crates/nh-remote/src/remote.rs
```diff
@@ -416,12 +415,28 @@ fn get_ssh_control_dir() -> &'static PathBuf {
...
+    }
+  }
+}
+
 #[derive(Eq, PartialEq, Debug, Clone)]
 pub struct RemoteHost {
   /// The host string (may include user@)
-  host: String,
+  host:         String,
+  store_scheme: NixStoreScheme,
```
NotAShelf, line 445: `RemoteHost::parse("host")` and `RemoteHost::parse("ssh://host")` seem to produce two `RemoteHost` values that technically represent the same physical machine. But what do you know, those won't compare equal with == because the derived `PartialEq` includes `store_scheme`. `"host"` gets `SshNg` and `"ssh://host"` gets `Ssh`. All existing comparisons go through `.hostname()` anyway so this has no big impact (therefore this is a nit rather than a blocker) but it's a potential footgun for future stuff we might implement. Since we're working on cleaning up nh I'd consider either manually implementing `PartialEq` to compare only the host field, or adding a test/doc comment to remind ourselves that this is quite intentional.

crates/nh-remote/src/copy.rs
NotAShelf: Then I'm keeping this comment open for now.

NotAShelf (approved):

PR #751, opened 2026-07-26
nh-search: try backend-version `n+1` when current version is unavailable

NotAShelf:
Adds support for searching and displaying nix-darwin options in `nh seach` and updates the search logic to handle a single fallback attempt (n+1) when the hardcoded backend version is unavailable. It tries only the next version to avoid stalling the search for too long.


crates/nh-search/src/args.rs
```diff
@@ -5,6 +5,7 @@ use color_eyre::{Result, eyre::bail};
 
 const DEFAULT_LIMIT: u64 = 30;
 const DEFAULT_CHANNEL: &str = "nixos-unstable";
+const DEFAULT_BACKEND_FALLBACKS: u32 = 1;
```
faukah, line 8: Nit: something more like this would be a bit better. 
```suggestion
const DEFAULT_BACKEND_FALLBACK_COUNT: u32 = 1;
```

NotAShelf: I don't think FALLBACKS is that ambiguous.

faukah (approved):

PR #579, opened 2026-02-27
Use the target hostname to determine the correct nixosConfiguration for remote builds

Stebalien:
The first patch in this series is a small "prefactor" that parses `RemoteHost` arguments inside the argument parser itself instead of at the use-sites (simplifies error handling, avoids duplicate code, etc).

The second patch changes the `nh os` commands to deploy the configuration named by the target hostname instead of the local hostname for remote deployments, except when the `-H|--hostname` flag is passed.

Tested by deploying a local config and a remote config.

fixes #572

## Sanity Checking

[changelog]: https://github.com/nix-community/nh/tree/master/CHANGELOG.md

- [x] I have updated the [changelog] as per my changes
- [x] I have tested, and self-reviewed my code
- Style and consistency
  - [x] I ran **`nix fmt`** to format my Nix code
  - [x] I ran **`cargo fmt`** to format my Rust code
  - [x] I have added appropriate documentation to new code
  - [x] My changes are consistent with the rest of the codebase
- Correctness
  - [x] I ran **`cargo clippy`** and fixed any new linter warnings.
- If new changes are particularly complex:
  - [ ] My code includes comments in particularly complex areas to explain the
        logic
  - [x] I have documented the motive for those changes in the PR body or commit
        description.
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/nix-community/nh/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

## Summary by CodeRabbit

* **Changes**
  * Remote deployment host selection now defaults to the target host unless a hostname is explicitly provided, simplifying remote workflows.
  * Host handling and specification improved: standardized parsing, richer host representation, and clearer hostname resolution when not provided.
  * SSH/remote operations made more robust with refined host validation and consistency across build and activation flows.

PR #771, opened 2026-08-15
Strip `.local` TLD from hostname for use with mDNS

LucasOe:
Follow up for #579.

With this PR `.local` suffixes get stripped from the hostname so that IPs using Multicast DNS resolve correctly. E.g. `user@host.local` will now use the hostname `host` instead of `host.local`.

The user can now deploy using mDNS by running `nh os switch --target-host user@host.local` without having to pass `--hostname host` in addition.

## Sanity Checking

- [x] I have read and understood the [contribution guidelines]
- Style and consistency
- Correctness
  - [ ] I ran **`cargo clippy`** and fixed any new linter warnings.
        logic
        description.
- Tested on platform(s):
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ be put in the "Changed" section or, if it's just to remove code or
 functionality, under the "Removed" section.
 -->
 
+## Unreleased
+
+### Added
+
+- `.local` TLDs now get stripped from the hostname. `user@host.local` now uses
+  `host` as a hostname instead of `host.local` for target and build hosts.
+
 ## 4.4.2
 
 ### Added
--- a/crates/nh-remote/src/remote.rs
+++ b/crates/nh-remote/src/remote.rs
@@ -488,14 +488,15 @@ pub struct RemoteHost {
 }
 
 impl RemoteHost {
-  /// Get the hostname part without the `user@` prefix.
+  /// Get the hostname part without the `user@` prefix and `.local` TLD.
   ///
   /// Used for hostname comparisons when determining if two `RemoteHost`
   /// instances refer to the same physical host (e.g., detecting when
   /// `build_host` == `target_host` regardless of different user credentials).
   ///
   /// Returns the bracketed IPv6 address as-is if present (e.g.,
-  /// `[2001:db8::1]`).
+  /// `[2001:db8::1]`). A `.local` suffix is stripped if present, so that
+  /// `user@host.local` returns `host`.
   ///
   /// # Panics
   ///
@@ -505,7 +506,8 @@ impl RemoteHost {
   #[must_use]
   pub fn hostname(&self) -> &str {
     #[allow(clippy::unwrap_used)]
-    self.host.rsplit('@').next().unwrap()
+    let hostname = self.host.rsplit('@').next().unwrap();
+    hostname.strip_suffix(".local").unwrap_or(hostname)
   }
 
   /// Parse a host specification string.


teto:
> user@host.local will now use the hostname host instead of host.local.

isn't user@host.local more explicit ? how is the machine supposed to find "host" without the ".local" (apart from ssh shenaningans) ? or in absence of FQDN it falls back to `.local` ? I am not sure ".local" is a standard, I've seen ".home" in the wild as well.

LucasOe:
> how is the machine supposed to find "host" without the ".local" (apart from ssh shenaningans)

The idea was that only the hostname (without the TLD) is used to get the correct flake output when inferring the name from `--target-host`. That means running `nh os switch --target-host user@foo.local` should work without specifying `--hostname foo` in addition.

I didn't notice that the [ssh_host](https://github.com/nix-community/nh/blob/0556d89ba6d63df9d423f3e0df8450d6653a0aae/crates/nh-remote/src/remote.rs#L614) function also uses the hostname to resolve the ssh connection.

I must have really messed up during testing. My bad.

> I am not sure ".local" is a standard

`.local` is indeed the standard as mentioned in the [RFC](https://www.rfc-editor.org/info/rfc6762/#appendix-G).

teto:
> .local is indeed the standard as mentioned in the [RFC](https://www.rfc-editor.org/info/rfc6762/#appendix-G).

thanks for the piece of information.

I didnt think about the UX improvement, was working on a nixpkgs avahi PR, looking at this repo for fun when I noticed this PR and ended up commenting as a passerby. Keep the good work ;)

LucasOe:
I have updated the changelog, documentation and PR body to hopefully make the intended UX improvement more clear. I also added tests.

Connections are still resolved using the Fully qualified domain name (FQDN), but the target host now strips `.local` TLDs when selecting the `nixosConfiguration`.

I decided to only strip `.local` TLDs instead of _all_ TLDs because hostnames can include dots as far as I know. E.g. `host.example.com` should be a valid hostname, with `host.example.com.local` used when connecting to it over mDNS.

NotAShelf (approved):
LGTM.

CHANGELOG entry arguably belongs under "Fixed" rather than "Added" since it closes #579 but it's not a dealbreaker. Technically both.

LucasOe:
I fixed the formatting and moved the changelog entry.

NotAShelf:
Also two nits, neither are blockers:

1. A flake entry literally named `host.local` (non-mDNS) would now resolve to host and fail. Probably not likely.
2. `host.LOCAL` (uppercase) would connect via SSH but not get its suffix stripped due to case-sensitive matching. Probably fine as is.

Up to you whether you want to address my comments. I'll merge this as is otherwise.

LucasOe:
> A flake entry literally named `host.local` (non-mDNS) would now resolve to host and fail. Probably not likely.

Yeah, I thought about this but I can't think of a good way to avoid it. Since it's very unlikely to happen, I think it's fine if a user has to pass `--hostname host.local` in this case.

> `host.LOCAL` (uppercase) would connect via SSH but not get its suffix stripped due to case-sensitive matching. Probably fine as is.

I don't know if there's a nice way to make `.stip` case insensitive. I looked at a few solutions, but they were all really ugly, haha. I think you could come up with something better than me if you want this fixed. If you want me to change this anyway, then let me know.

NotAShelf:
Thanks :)

PR #793, opened 2026-09-12
fix(remote): pass --no-link with remote --print-out-paths

rriski:
## Why

`nix build` creates `./result` unless `--no-link` is passed. `--print-out-paths` does not imply `--no-link`.

When `--build-host` equals `--target-host`, nh skips the local `--out-link` and leaves the result on the remote. A remote `nix build <drv>^* --print-out-paths` then fails for a system user whose home is `/var/empty`.

## What changed

Remote `nix build` always includes `--no-link` next to `--print-out-paths`. Extra `-- --no-link` from the caller stays valid.

Made with [Cursor](https://cursor.com)


faukah:
Please read [CONTRIBUTING.md](https://github.com/nix-community/nh/blob/master/CONTRIBUTING.md) before creating a PR in nh. LLM-generated contributions are generally not welcomed, especially if they completely fail to comply with the contribution guidelines.

PR #801, opened 2026-09-19
nh-darwin: activate without `darwin-rebuild`

### LINKED ISSUE #800: Cannot disable `darwin-rebuild` in nix-darwin config
### I have confirmed that this is a bug related to nh

- [x] This is a bug, and not an user error or a support request. I understand that my issue will be closed if it is not a bug in NH.
- [x] I have checked the [issues tab](https://github.com/nix-community/nh/issues?q=is%3Aissue) and confirmed that my issue has not yet been reported. I understand that my issue will be closed if it is a duplicate.

### Description

```
CHANGED
[C-] darwin-rebuild <none> ×2 -> <none>, -10.4 KiB

PATHS: 648 -> 647 (+7, -8)
SIZE: 8.84 GiB -> 8.84 GiB
DIFF: -11.0 KiB
> Apply the config? Yes
Password:
> Activating configuration
Error:
   0: Darwin activation failed
   1: Activating configuration (exit status ExitStatus(Exited(127)))
      stderr:
      env: ‘/var/folders/ps/6cck97391xx1m966xtcmg0l80000gn/T/nh-darwinoUTJsD/result/sw/bin/darwin-rebuild’: No such file or directory


Location:
   crates/nh-core/src/command.rs:931
```

https://github.com/nix-community/nh/blob/3c05c012c090a20c5a8e6501e756a6109c86c459/crates/nh-darwin/src/darwin.rs#L203-L224

darwin-rebuild succeeds though so this is nh-specific

### Installation Method

Flakes (`nix profile add`) / Channels (`nix-env` from master branch or from a release)

### Installation Method (Other)

_No response_

### NH Version

4.4.2

### System Information

aarch64-darwin

NotAShelf:
Refactors the activation logic in the Darwin rebuild process to better support both modern and legacy activation workflows *without* `darwin-rebuild`. We now explicitly handle the presence of a non-deprecated `activate-user` script as an indicator for legacy user activation


crates/nh-darwin/src/darwin.rs
```diff
@@ -200,22 +200,32 @@ impl DarwinRebuildArgs {
         .run()
         .wrap_err("Failed to set Darwin system profile")?;
 
-      let darwin_rebuild = out_path.join("sw/bin/darwin-rebuild");
+      let activate = out_path.join("activate");
       let activate_user = out_path.join("activate-user");
```
eclairevoyant, line 204: activate-user should be removed, as it was deprecated before 25.11:

https://github.com/nix-darwin/nix-darwin/blob/4cff07de74b50e64bdd68cd4e722ab5b6b35ee48/modules/system/default.nix#L108-L133

eclairevoyant (approved):
Approving since it fixes the issue

crates/nh-darwin/src/darwin.rs
NotAShelf: It's strictly a legacy compatibility path. I don't use Darwin so I h ave to ask: are you certain we can absolutely drop this without affecting users?

eclairevoyant: This will affect users that can't read a warning message lol. It's going to be removed upstream anyway:

* https://github.com/nix-darwin/nix-darwin/pull/1825

eclairevoyant (approved):

